PR #6, opened 2021-01-12
Issue/none introduce connection state

wyrzyk:


src/main/java/com/atlassian/db/replica/api/state/States.java
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.db.replica.api.state;
+
+public enum States {
```
wyrzyk, line 3: I probably should not use enums for API.

dagguh: Hmm, but transitions would be tricky: `getReadConnection()` would have to return both a `Connection` and a new `ConnectionState`. Or become mutable.

src/main/java/com/atlassian/db/replica/spi/state/StateListener.java
```diff
@@ -0,0 +1,7 @@
+package com.atlassian.db.replica.spi.state;
+
+import com.atlassian.db.replica.api.state.States;
+
+public interface StateListener {
```
dagguh, line 5: So far it seems it only has a no-op impl. What's the target use case?

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -0,0 +1,178 @@
...
+        if (isClosed != null && isClosed) {
+            return CLOSED;
+        } else if (!readConnection.isInitialized() && !writeConnection.isInitialized()) {
+            return NOT_INITIALISED;
+        } else if (writeConnection.isInitialized()) {
+            return MAIN;
+        } else if (readConnection.isInitialized()) {
+            return REPLICA;
+        } else {
+            throw new UnsupportedStateException();
```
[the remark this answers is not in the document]
wyrzyk: The whole class and DualConnection is not thread-safe ;)

src/main/java/com/atlassian/db/replica/spi/state/StateListener.java
wyrzyk: 1. It's easier to test how states changes.
2. We can track transitions more gradually (for example, with the stack-trace mechanism) form the app.

I'm not 100% sure if we need this extra visibility available in API.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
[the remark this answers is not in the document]
dagguh: Yeah but it didn't throw `UnsupportedStateException`s earlier.
Perhaps the race condition is bogus, if we have the guarantee that `isInitialized` cannot go from `true` to `false`.

src/main/java/com/atlassian/db/replica/api/state/States.java
wyrzyk: What do you mean? Can you express it with a test case?

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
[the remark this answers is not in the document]
wyrzyk: There's no guarantee. 

- [ ] Avoid `UnsupportedStateExceptions`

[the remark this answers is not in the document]
dagguh: Right, we have the `.reset`

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -1,104 +1,75 @@
...
-        initializedConnections.clear();
-        if (!autoCommitBefore.equals(getAutoCommit())) {
+    public void setAutoCommit(Boolean autoCommit) throws SQLException {
+        final boolean autoCommitBefore = getAutoCommit();
+        parameters.setAutoCommit(autoCommit);
+        final Optional<Connection> connection = state.getConnection();
+        if (connection.isPresent()) {
+            connection.get().setAutoCommit(autoCommit);
+        }
+        if (autoCommitBefore != getAutoCommit()) {
```
dwatl, line 62: autoCommitPrev(ious)?

dwatl (approved):
LGTM

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
wyrzyk: Can you elaborate?

dwatl: I meant the naming, but it's highly subjective - **previous** seemed more clear than **before**, but I'm not sure now

src/main/java/com/atlassian/db/replica/api/state/State.java
```diff
@@ -0,0 +1,37 @@
...
+    private final String name;
+
+    private State(String name) {
+        this.name = name;
+    }
+
+    public static final State NOT_INITIALISED = new State("NOT_INITIALISED");
+    public static final State MAIN = new State("MAIN");
+    public static final State REPLICA = new State("REPLICA");
+    public static final State CLOSED = new State("CLOSED");
```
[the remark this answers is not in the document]
dagguh: Oops, I forgot to submit the review 😅 
Either way, merging is a good call. Due to VPN problems I didn't find time to prepare the example, so better merge than wait 👍

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -1,104 +1,75 @@
...
-        final Boolean autoCommitBefore = getAutoCommit();
-        this.isAutoCommit = autoCommit;
-        initializedConnections.clear();
-        if (!autoCommitBefore.equals(getAutoCommit())) {
+    public void setAutoCommit(Boolean autoCommit) throws SQLException {
+        final boolean autoCommitBefore = getAutoCommit();
+        parameters.setAutoCommit(autoCommit);
+        final Optional<Connection> connection = state.getConnection();
+        if (connection.isPresent()) {
+            connection.get().setAutoCommit(autoCommit);
```
dagguh, line 60: Can't use `Optional.ifPresent` due to `throws SQLException`
Checked exceptions trying to help again 😂

PR #11, opened 2021-01-26
Generalize circuit breaker SPI

dagguh:
`CircuitBreaker` SPI doesn't enumerate states anymore.
It only asks if it can be called and how to handle calls.

This lets you track any state and wire in advanced CircuitBreaker libs.
For example, `resilience4j` defines more exotic states like `DISABLED` or
or `FORCED_OPEN`.
It also counts duration of calls and various other features.

So instead of reimplementing all possible circuit breaker features,
the SPI only asks for methods, which `db-replica` actually needs.


src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakOnNotSupportedOperations.java
```diff
@@ -1,30 +1,42 @@
...
 public class BreakOnNotSupportedOperations implements CircuitBreaker {
-    private static volatile BreakerState state = CLOSED;
+
+    private static volatile boolean LACK_OF_SUPPORT_DETECTED = false;
 
     @Override
-    public BreakerState getState() {
-        return state;
+    public boolean canCall() {
+        return !LACK_OF_SUPPORT_DETECTED;
```
dagguh, line 16: I'll get rid of double negation

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakerHandler.java
```diff
@@ -1,36 +0,0 @@
...
-    private final CircuitBreaker breaker;
-
-    public BreakerHandler(CircuitBreaker breaker) {
-        this.breaker = breaker;
-    }
-
-    public <T> T handle(SqlCall<T> call) throws SQLException {
-        try {
-            return call.call();
-        } catch (Throwable throwable) {
```
dagguh, line 18: Bonus: we no longer catch `Throwable`, which is [a smell](https://rules.sonarsource.com/java/RSPEC-1181):

> Catching either `Throwable` or `Error` will also catch `OutOfMemoryError` and `InternalError`, from which an application should not attempt to recover.

If a `CircuitBreaker` impl chooses to catch one anyway, it's on them, not forced by `db-replica`.

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
```diff
@@ -1,9 +1,18 @@
 package com.atlassian.db.replica.spi.circuitbreaker;
 
-import com.atlassian.db.replica.api.circuitbreaker.BreakerState;
```
dagguh, line 3: `CircuitBreaker` is the only member of `spi.circuitbreaker`. The `api.circuitbreaker` doesn't even exist. Shall I move it to `spi`?

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakOnNotSupportedOperations.java
```diff
@@ -1,30 +1,42 @@
...
+            return call.call();
+        } catch (ReadReplicaUnsupportedOperationException | SQLFeatureNotSupportedException e) {
+            ALL_CALLS_SUPPORTED_SO_FAR = false;
+            throw e;
+        }
     }
 
     @Override
-    public void handle(Throwable throwable) {
-        if (throwable instanceof ReadReplicaUnsupportedOperationException || throwable instanceof SQLFeatureNotSupportedException) {
```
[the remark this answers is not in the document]
wyrzyk: But not without a cost. At the and we always care to handle throwable. The current implementation requires to implement two methods and may grow in the future. The new interface is also more error-prone. The wrong implementation can cut all the calls to DB.

+ no `instanceof`
- internal implementation promotes code duplication,
- harder to implement,
- less stable SPI,
- broader SPI,
- more error-prone,

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakOnNotSupportedOperations.java
```diff
@@ -1,30 +1,42 @@
...
+import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 
-import static com.atlassian.db.replica.api.circuitbreaker.BreakerState.CLOSED;
-import static com.atlassian.db.replica.api.circuitbreaker.BreakerState.OPEN;
-
 public class BreakOnNotSupportedOperations implements CircuitBreaker {
-    private static volatile BreakerState state = CLOSED;
+
+    private static volatile boolean ALL_CALLS_SUPPORTED_SO_FAR = true;
```
wyrzyk, line 12: Nit. We can re-use cache here. This way, it could be easily extended to use external cache and work across nodes.

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakerHandler.java
wyrzyk: We catch and rethrow it, so the smell is not relevant.

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
```diff
@@ -1,9 +1,18 @@
...
-    BreakerState getState();
 
-    void handle(Throwable throwable);
+    boolean canCall();
+
+    <T> T handle(SqlCall<T> call) throws SQLException;
+
+    void handle(SqlRunnable runnable) throws SQLException;
+
+    interface SqlRunnable {
```
wyrzyk, line 15: Why inner? Doesn't deserve their own file?

wyrzyk: BTW why do we need it?

wyrzyk:
> `CircuitBreaker` SPI doesn't enumerate states anymore.
> It only asks if it can be called and how to handle calls.
> 
> This lets you track any state and wire in advanced CircuitBreaker libs.
> For example, `resilience4j` defines more exotic states like `DISABLED` or
> or `FORCED_OPEN`.
> It also counts duration of calls and various other features.
> 
> So instead of reimplementing all possible circuit breaker features,
> the SPI only asks for methods, which `db-replica` actually needs.

I like we get rid of the state. It makes SPI easier, more powerful and more stable.

I don't like the handler, as it makes SPI worse without any benefit for the end-user.

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
```diff
@@ -1,9 +1,18 @@
...
+
+import java.sql.SQLException;
 
 public interface CircuitBreaker {
-    BreakerState getState();
 
-    void handle(Throwable throwable);
+    boolean canCall();
+
+    <T> T handle(SqlCall<T> call) throws SQLException;
```
wyrzyk, line 11: AFAIK we can use `DualCall` to handle all calls to the DB. Maybe,  `CircuitBreaker` interface can contain only `boolean canCall();` method?

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
```diff
@@ -1,9 +1,18 @@
...
-import com.atlassian.db.replica.api.circuitbreaker.BreakerState;
+import com.atlassian.db.replica.api.SqlCall;
+
+import java.sql.SQLException;
 
 public interface CircuitBreaker {
-    BreakerState getState();
 
-    void handle(Throwable throwable);
+    boolean canCall();
```
wyrzyk, line 9: It may be misleading. We don't call this method when we decide if we want to call DB or not. It's rather `canEstablishConnectionToReplica`. 🤔

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakOnNotSupportedOperations.java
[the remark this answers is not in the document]
dagguh: In general, there are more benefits than you listed. Most are in the commit message: f07ba86bc8082f7d2c68cf6d465cbad416a8ec2c

> (-) internal implementation promotes code duplication,

That's also visible in `resilience4j` and anything with callbacks. The `Runnable`/`Callable` split is standard in Java.

> (-) harder to implement,

True! Unless you want to add latency-based CB, or other states, then it's easier.

> (-) less stable SPI,
> (-) broader SPI,

Circuit breakers can have a lot of methods. In `resilience4j` it has 40 instance methods and 14 static methods. Here I just chose a subset (3) that is actually needed by `db-replica`. So it is as narrow as possible. Whatever CB feature is needed by Jira, it is possible with this SPI, so there's less reason to change it, so it's more stable. It can be less stable if `db-replica` needs more from the CB, but that remains to be seen.

> (-) more error-prone,

Old impl can also cut all calls to DB, e.g. by returning the wrong state.

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakerHandler.java
dagguh: We catch, pass to SPI and then rethrow it. We dunno what happens in the SPI.
But it's not the point of the change, just a "bonus". Even if all impls behave, then at least SonarLint stops the warnings 😉

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakOnNotSupportedOperations.java
dagguh: Oooh, good point

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
dagguh: It's moved from the old inner `BreakerHandler.SqlRunnable`. I will extract it.

I will also try to remove the need for it

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
dagguh: Actually, it's `canUseDualConnection`

wyrzyk: It will still use DualConnection's that are in the fly. It will not create new ones. `canCreateDualConnection`?

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
dagguh: Cannot get rid of `SqlRunnable`:
image

Standard split is `Callable`/`Runnable`.
Now we have `SqlCall`/`SqlRunnable`.
For consistency, we should go for:
1. short: `SqlCall`/`SqlRun`
2. idiomatic: `SqlCallable`/`SqlRunnable`

I'd default to *short*, but I will not defend that

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakOnNotSupportedOperations.java
dagguh: Done

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
dagguh: `DualCall` is about splitting behavior across main/replica
`CircuitBreaker` is about protecting any usage of `DualConnection`, regardless of main/replica

wyrzyk: `DualCall` is about intercepting calls to main and replica. It can be used to add logs, monitoring or feed CircuitBreaker with better granularity than the current SPI of CB ("entities should not be multiplied without necessity"). It probably can and will be used in the future for use cases we can't imagine now.

dagguh: `BreakerConnection` doesn't use it. Should it? But `BreakerConnection` doesn't split know about main/replica.

wyrzyk: > BreakerConnection doesn't use it. Should it?

IMO it should use it. `BreakerConnection` was created with `MissingCoverageBreaker` implementation in mind, which doesn't need the split. But it's a special case. In general, you probably want to detect when the replica fails. If only connections to the main are failing, then there's no reason to switch all the traffic to the main database.

If we add an extra entity (CB#handle), it will be more limited than the current SPI `DualCall`.

dagguh: I understand

- [ ] remake `BreakerConnection` to handle replica calls only

wyrzyk: I'm not sure if we are on the same page. `BreakerConnection` can handle both calls,  how to implement `MissingCoverageBreaker` with replica calls only?

wyrzyk: Currently, we have two ways to do the same. One gives us a more granular view of what happens (splits calls), another gives us only an aggregation. I'd leave more powerful way instead of:
- adding an alternative way to do the same but with limits (aggregation),
- limiting CB only to get part of the calls.

src/main/java/com/atlassian/db/replica/api/MissingCoverageBreaker.java
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            run.run();
+        } catch (ReadReplicaUnsupportedOperationException | SQLFeatureNotSupportedException e) {
+            throw handleMissingCoverage(e);
+        }
+    }
+
+    private RuntimeException handleMissingCoverage(Exception e) {
+        coverageSufficient.put(false);
+        String ticket = "https://github.com/atlassian-labs/db-replica/issues/new";
```
tciesla-atlassian, line 55: Shouldn't it be moved to constant (private static final field)?

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakerConnection.java
```diff
@@ -22,121 +24,121 @@
 
 public class BreakerConnection implements Connection {
     private final Connection delegate;
-    private final BreakerHandler breakerHandler;
+    private final CircuitBreaker breaker;
```
tciesla-atlassian, line 27: It looks weird when instance fields is called breaker, does it break something? Full name circutBreaker looks better ;)

src/main/java/com/atlassian/db/replica/spi/circuitbreaker/CircuitBreaker.java
wyrzyk: Please check this PR https://github.com/dagguh/db-replica-1/pull/1

dagguh:
Thanks for the comments folks! I'll catch up tomorrow

src/main/java/com/atlassian/db/replica/api/MissingCoverageBreaker.java
dagguh: What's the purpose?
If performance, then String literals are interned in the String pool in heap, so it won't make a difference.
From readability angle, keeping things local is easier on the brain.

src/main/java/com/atlassian/db/replica/internal/circuitbreaker/BreakerConnection.java
dagguh: Fair point

dagguh:
If I understand correctly, [one way](https://github.com/dagguh/db-replica/pull/1#issuecomment-785157972) is to remove `CircuitBreaker` SPI entirely and let consumers handle that via their own `DatabaseCall` aspects. Is that the idea @wyrzyk?

dagguh:
#60 removed CB from API, so reshaping it is no longer necessary

Issue #10, opened 2021-01-26
Try to switch to replica after inconsistency


wyrzyk:
Currently, [MainConnection is a final state](https://github.com/atlassian-labs/db-replica/blob/master/docs/dual-connection-states.md). When `DualConnection` starts to use a connection to the main database, it will always use this connection (until it returns to the connection pool).

One of the reasons to use the main database is inconsistency. If we switch to the main connection because of the inconsistency, it should be possible to switch back to the replica.

PR #12, opened 2021-01-28
#10: Try to switch to replica after inconsistency

wyrzyk:


docs/dual-connection-states.md
```diff
@@ -13,8 +13,24 @@ the replica is ready to serve consistent responses.
...
 
+## States
+
+- `NoConnection` - when we create a new `DualConnection` it doesn't allocate
+  any real database connection.
+- `ReplicaConnection` - connection to the replica database has been established. It will be
+  re-used until we need to switch to either `MainConnection` or `ReadFromMainConnection`.
+- `MainConnection` - connection to the main database has been established and used to do
+  writes to the database. `DualConnection` will keep using it until closed.
+  The main reasons not to switch to replica from this state:
```
dagguh, line 28: Explicit rationale is great!

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -22,6 +26,7 @@
     private final ConnectionParameters parameters;
     private final Warnings warnings;
     private final StateListener stateListener;
+    private volatile boolean isReplicaInconsistent = false;
```
dagguh, line 29: Nits:
* drop the `is` (it's for a getter)
* remove negation: `boolean replicaConsistent`

docs/dual-connection-states.md
```diff
@@ -13,8 +13,24 @@ the replica is ready to serve consistent responses.
...
+    - to avoid affecting transactions
+    - to avoid affecting locks
+- `ReadFromMainConnection` - connection to the main database has been established, but only
```
dagguh, line 31: This one is new. It's missing from the UML

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
```diff
@@ -163,4 +165,87 @@ public void shouldChangeStateFromMainToClosed() throws SQLException {
...
+            connectionProvider,
+            new CircularConsistency.Builder(ImmutableList.of(false, true)).build()
+        ).stateListener(stateListener)
+            .build();
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+        Mockito.reset(stateListener);
+
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+
+        verify(stateListener).transition(READ_FROM_MAIN, REPLICA);
```
dagguh, line 214: So this is the whole point of the change.
I don't understand why a `transition(MAIN, REPLICA)` wasn't allowed before. Is there a test showing that it would be a problem?

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
```diff
@@ -163,4 +165,87 @@ public void shouldChangeStateFromMainToClosed() throws SQLException {
...
+    }
+
+    @Test
+    public void shouldChangeStateFromReadFromMainConnectionToMainConnectionConnection() throws SQLException {
```
dagguh, line 218: Typo
```suggestion
    public void shouldChangeStateFromReadFromMainConnectionToMainConnection() throws SQLException {
```

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
```diff
@@ -163,4 +165,87 @@ public void shouldChangeStateFromMainToClosed() throws SQLException {
...
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+        Mockito.reset(stateListener);
+
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+
+        verify(stateListener).transition(REPLICA, READ_FROM_MAIN);
+    }
+
+    @Test
+    public void shouldChangeStateFromReadFromMainConnectionToReplicaConnection() throws SQLException {
```
dagguh, line 201: This redescribes the assertion. But why or when should it change the state in that way?

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
```diff
@@ -163,4 +165,87 @@ public void shouldChangeStateFromMainToClosed() throws SQLException {
...
+        final StateListener stateListener = mock(StateListener.class);
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
```
dagguh, line 203: Can we put those in `@Before` to unclutter the test cases?

docs/dual-connection-states.md
wyrzyk: > It's missing from the UML

What do you mean? The UML is updated.

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
wyrzyk: It was not implemented before.

docs/dual-connection-states.md
dagguh: My, bad, I viewed rich-diff of the `.md`, but I should look at diff of `.puml` and `.png`.

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
dagguh: Can we reuse `MAIN` instead of `READ_FROM_MAIN`? If not, why? I guess:
* avoid replica forever after any write for the rest of the connection life (but why so long?)
* avoid replica only when inconsistent (makes sense, it's "forgiving")

Why are writes so unforgiving? Because read-after-write? But that's what `ReplicaConsistency` is for. Is it about transactions?

wyrzyk: > Can we reuse MAIN instead of READ_FROM_MAIN?

MAIN and READ_FROM_MAIN hold information about a state. If we want to re-user MAIN state, we need to store part of the state somewhere else. 

> Why are writes so unforgiving? 

- transactions
- locks
- transaction isolation level

> But that's what ReplicaConsistency is for.

`ReplicaConsistency` works only after the changes are committed and visible outside of the current transaction.
When auto-commit == false, we can't release the connection until the transaction is committed manually. 

The initial implementation was a POC, and to deliver it faster, some shortcuts needed to be made. One of them was to avoid switching to a replica from the main. It cut lots of complexity, but it's less effective. Now we can incrementally add support for some transitions. The simplest one is when we used the main database connection because of the inconsistency.

Maybe `READ_FROM_MAIN` state can be more generic. If we decide to `transition(MAIN, REPLICA)` after a transaction committed, we could re-use the state. 🤔  My propositions:
- CLEAR/CLEAN_MAIN - the main database is in the state that allows us to `transition(MAIN, REPLICA)`. Main connection's state is clear (not-dirty)
- COMMITED_MAIN - all the changes from this connection are already committed, we can rely on ReplicaConsistency. We can drop the connection without worrying about affecting current transaction/locks 🤷 .

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
wyrzyk: `shouldChangeStateFromReadFromMainConnectionToReplicaConnectionWhenNoWritesOnMain` will be better?

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
wyrzyk: For me, it's easier to read code when there's no mutable class lever variable, but I can add before if you like it more. I have no strong feelings.

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
dagguh: Nice! Now I know the what **and** the why.
Lemme try some rewording iterations:
* firstly, drop redundancy: `shouldSwitchFromMainToReplicaWhenNoWritesOnMain`
* secondly some semantic compression: `shouldForgiveReplicaWhenNoWritesOnMain`
* thirdly, reducing negation:  `shouldForgiveReplicaIfItCatchesUpOnReads`.

And counterpart test case (if it exists): `shouldAvoidReplicaDueToPossibleTransactions`.

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
dagguh: Fair enough, for now let's optimize for state tracking. I have some more ideas on how to optimize for readability, so it will be a better ROI later.

src/test/java/com/atlassian/db/replica/internal/state/ConnectionStateTest.java
dagguh: I see! So `MAIN` "tracks" transactions/locks and `READ_FROM_MAIN` tracks only the inconsistencies. And `MAIN` could be optimized later to also be forgiving. Thanks for the context! It'd be nice to have the word "transaction" pop up somewhere in the test case.

docs/dual-connection-states.md
```diff
@@ -13,8 +13,24 @@ the replica is ready to serve consistent responses.
...
+  re-used until we need to switch to either `MainConnection` or `CommitedMain`.
+- `MainConnection` - connection to the main database has been established and used to do
+  writes to the database. `DualConnection` will keep using it until closed.
+  The main reasons not to switch to replica from this state:
+    - to avoid affecting transactions
+    - to avoid affecting locks
+- `CommitedMain` - connection to the main database has been established, but only
+  for reads. The replica was temporarily inconsistent. This state shares some features
+  of both `ReplicaConnection` and `MainConnection` states. It's connected to the main database,
+  but can switch to the replica.
```
tciesla-atlassian, line 34: +1 nice explanation

src/main/java/com/atlassian/db/replica/api/state/State.java
```diff
@@ -14,6 +14,7 @@ private State(String name) {
     public static final State MAIN = new State("MAIN");
     public static final State REPLICA = new State("REPLICA");
     public static final State CLOSED = new State("CLOSED");
+    public static final State COMMITED_MAIN = new State("COMMITED_MAIN");
```
tciesla-atlassian, line 17: BTW why this is class, not "enum"?

tciesla-atlassian (approved):

src/main/java/com/atlassian/db/replica/api/state/State.java
dagguh: [Enums are hard to evolve](https://github.com/wyrzyk/java-api-design-checklist/blob/a63b78efe8a988343ecdb7bfaf3f81079f5ebae2/CHECKLIST.md#255-do-not-add-new-values-to-a-released-enumeration)

wyrzyk: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html
> You should use enum types any time you need to represent a fixed set of constants. That includes natural enum types such as the planets in our solar system and data sets where you know all possible values at compile time—for example, the choices on a menu, command line flags, and so on.

We don't know the fixed set of states yet. Using enum would block [the API's evolution](https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#255-do-not-add-new-values-to-a-released-enumeration).

dagguh, line 17: Do we need to expose those `public static final State COMMITED_MAIN` and its siblings?

wyrzyk: I guess it's easier to compare classes than.
Currently, I'm working on a query level visibility, and I'm not sure if it will not deprecate the current STATE level visibility changes. I'll postpone the merge and see how it evolves.

wyrzyk: I'll merge it and resolve this issue before 1.0.0 release.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -513,9 +515,10 @@ void recordWriteAfterQueryExecution() throws SQLException {
 
     public Statement getReadStatement(String sql) {
         isWriteOperation = isFunctionCall(sql) || isUpdate(sql) || isDelete(sql);
+
         // if write connection is already initialized, but the current statement is null
         // we should use a write statement, regardless of the fact that readStatement has been called.
-        if (connectionProvider.hasWriteConnection() || isWriteOperation || isSelectForUpdate(sql)) {
+        if (connectionProvider.getState().equals(MAIN) || isWriteOperation || isSelectForUpdate(sql)) {
```
afaruga-atlassian, line 521: Right now `ReplicaStatement` need to know about MAIN, what increase coupling, and I think it would be better to have two different methods

1. `hasWriteConnection`
2. `hasOnlyWriteConnection`

wyrzyk: > Right now ReplicaStatement need to know about MAIN

The state is part of the API. What's wrong in `knowing` it?

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -101,6 +108,7 @@ public Connection getReadConnection() throws SQLException {
      */
     public Connection getWriteConnection() throws SQLException {
         final State stateBefore = getState();
+        replicaConsistent = true;
```
afaruga-atlassian, line 111: This looks weird to me.
Like you getting connection - it's got the operation and should be independent, as it won't make it pure function.
Maybe this is expected, but looking at this for the first time looks like a design smell and I've thought I'll share that thinking with you.

For me, this might be a suggestion to expand the logic behind replicaConsistent and doesn't have it as a single boolean value.

PR #13, opened 2021-02-03
JCES-1757: Add docs and diagrams

wyrzyk:
`read-replica` had poor documentation. It was hard to understand it on a high level.
I hope it's enough to lower the barrier of entry for the libraries consumers and
for the libraries contributors.

I skipped details about `ReplicaConsistency` as it's currently under development and
will probably change soon.


docs/switching-between-main-and-replica.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. A [connection's state](dual-connection-states.md).
+2. A Replica's [consistency](../src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java).
+3. Context of `java.sql.Connection`/`java.sql.Statement` API usage.
+
+Some of the methods are intended to write into the database. For example every call to `java.sq.PreparedStatement#executeUpdate`
+will switch the connection's state to the main database.
+
+4. Availability of replica
+5. The query will use the main database in case it's:
```
dwatl, line 14: Should we mention something about nested/complex queries?

dwatl (approved):

docs/switching-between-main-and-replica.md
wyrzyk: What do you mean?

dwatl: This mentions simple cases only. I remember we are using "startsWith" to check if there is DELETE/UPDATE etc.
Can there be a query that starts with select, but it's doing update later on? e.g. https://stackoverflow.com/a/6764367

wyrzyk: Yes. Not all the cases are explained here. The tests are more detailed documentation, and in case you find a nice corner case I encourage you to contribute a test :). See https://github.com/atlassian-labs/db-replica/blob/19b83af9838413c414cb64c238b7ff819082394b/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java#L235, for example.

wyrzyk:
@dwatl Thanks for the review. I know you were familiar, I'm still waiting for a review from someone who never touched it.

dwatl (approved):

docs/classes-overview.md
```diff
@@ -0,0 +1,19 @@
+## Classes diagram
+
+This diagram is intended to help understand how the library works internally. It may leak some implementation details and should not be used as API.
```
afaruga-atlassian, line 3: @wyrzyk This is relly well-written description to well-written diagram. 
Helps dive into the project.
I'm sure it'll pay off long term to make a decision, use common languages, and helps also other ppl contribute and maintain the library.
㊗️

dagguh: ㊗️ = congratulations? 😃

docs/consistency.puml
```diff
@@ -0,0 +1,21 @@
+@startuml
+
+@startuml
+
+class DualConnection
+
+interface ReplicaConsistency{
```
afaruga-atlassian, line 7: I wonder if build would faild if ReplicaConsistency SPI would change, but `consistency.puml` not 
@wyrzyk WDYT?

wyrzyk: It would be great to have such functionality, currently `consistency.puml` is not connected with the source and we need to take care of it manually.

docs/classes-overview.md
afaruga-atlassian: Screenshot 2021-02-19 at 12 24 58
jej, this is what github prompt me;P

docs/consistency.puml
dagguh: True, although we shouldn't change architecture that often.

https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html :
> Explicitly call-out architectural invariants. Often, important invariants are expressed as an *absence* of something, and it’s pretty hard to divine that from reading the code. Nothing in the model layer specifically doesn’t depend on the views.

It's about high-level boundaries, what should go where and which parts are connected. It's not about reflecting the codebase precisely.

afaruga-atlassian: > True, although we shouldn't change architecture that often.
Agree. As long as dual-connection isn't big unit/component/lib with > 15 contributors this will work.
Later ppl just come and do their best efforts, and might do some mess by accident.
With that  [fitness functions](https://www.thoughtworks.com/radar/techniques/architectural-fitness-function)
comes very healthy!


Ok I dig a little bit;
[archunit-plantuml](https://www.archunit.org/userguide/html/000_Index.html#_plantuml_component_diagrams_as_rules)
inspired by  [this googled issue](https://github.com/TNG/ArchUnit/issues/108)

It's not for now(all goods) but I wonder whether You considered such an approach for dual-connection lib? 
Asking as I wasn't involved in the early days of dual-connection growth so I'm trying to get my eyes around it.

wyrzyk: > It's not for now(all goods) but I wonder whether You considered such an approach for dual-connection lib?

No, but feel free to contribute.

> Asking as I wasn't involved in the early days of dual-connection growth

We are still in the early days. We didn't reach the 1.0.0 release yet. 😉

PR #15, opened 2021-02-05
None: Add `State#getName` method

wyrzyk:
The method will allow creating a separate metric per transition.


dwatl (approved):

src/main/java/com/atlassian/db/replica/api/state/State.java
```diff
@@ -15,6 +15,10 @@ private State(String name) {
     public static final State REPLICA = new State("REPLICA");
     public static final State CLOSED = new State("CLOSED");
 
+    public String getName() {
```
dagguh, line 18: So a `toString`, that isn't vulnerable to future new fields? Fair enough

dagguh (approved):

PR #16, opened 2021-02-05
None: Reduce DualConnectionPerfIT flakiness on GH actions

wyrzyk:
It seems that GH action's infrastructure is flakier and can process the
`micro-benchmark` slower than we expect. 3k/ms is still an acceptable
thruput.


wyrzyk:
An example of a flaky test: https://github.com/atlassian-labs/db-replica/pull/15/checks?check_run_id=1838813878

dagguh (approved):
LGTM

Issue #20, opened 2021-02-09
As a developer, I'd like to know why a query goes to a particular database


wyrzyk:
Currently, the library lacks transparency on a query level. This feature would help understand how the library works and find queries that can potentially be migrated to read replica.

PR #21, opened 2021-02-09
Issue #20: Add query level visibility

wyrzyk:


wyrzyk:
- [ ] Review if documentation is up to data
- [ ] Resolve TODO
- [ ] Reformat
- [ ] Consider removing StateListener

src/main/java/com/atlassian/db/replica/api/context/QueryContext.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public String toString() {
+        return "QueryContext{" +
+            "reason=" + reason +
+            ", sql='" + sql + '\'' +
+            ", runOnMain=" + runOnMain +
+            '}';
+    }
+
+    public static QueryContext.Builder builder(Reason reason, boolean runOnMain) {
```
dwatl, line 52: What is the reason if we run on replica?

wyrzyk: READ_OPERATION -> https://github.com/atlassian-labs/db-replica/pull/21/files#diff-5eca8f926943fd2aca9f64d64f90d2d14ed51191273382f5edd4812e49e3d2abR645

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -426,47 +472,77 @@ public long getLargeMaxRows() throws SQLException {
...
         checkClosed();
         final Statement statement = getWriteStatement();
-        return execute(() -> statement.executeLargeUpdate(sql));
+        return execute(
+            () -> statement.executeLargeUpdate(sql),
+            QueryContext.builder(
+                API_CALL,
+                true
+            ).sql(sql)
+                .build()
```
dagguh, line 494: Can we reduce line shredding? There's plenty of horizontal space

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -581,10 +666,6 @@ protected Statement createStatement(Connection connection) throws SQLException {
         }
     }
 
-    public boolean isReadOnly() {
```
dagguh, line 584: Neat

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -513,14 +589,23 @@ void recordWriteAfterQueryExecution() throws SQLException {
...
+        if (connectionProvider.getState().equals(MAIN)) {
+            contextBuilder.reason(MAIN_CONNECTION_REUSE).runOnMain(true);
+            return prepareWriteStatement();
+        }
 
-        // if write connection is already initialized, but the current statement is null
-        // we should use a write statement, regardless of the fact that readStatement has been called.
-        if (connectionProvider.getState().equals(MAIN) || isWriteOperation || isSelectForUpdate(sql)) {
+        if (isSelectForUpdate(sql)) {
+            contextBuilder.reason(LOCK).runOnMain(true);
```
dagguh, line 604: Does it mean that `SELECT ... FOR UPDATE` locks the row/table/whatever and that's why replica is not ok? Or is the `UPDATE` itself the reason? Or both? Maybe we need `FOR_UPDATE` reason?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -513,14 +589,23 @@ void recordWriteAfterQueryExecution() throws SQLException {
...
             return prepareWriteStatement();
         }
+
+        readStatement.setContextBuilder(contextBuilder);
```
dagguh, line 608: This is where it would use the distant `READ_OPERATION` de facto default

src/main/java/com/atlassian/db/replica/api/context/QueryContext.java
```diff
@@ -0,0 +1,87 @@
+package com.atlassian.db.replica.api.context;
+
+
+import java.util.Objects;
+
+public final class QueryContext {
```
dagguh, line 6: "Context" is mega vague. Shall it hold *the* context of a query? But context is in the eye of the beholder. For some, "the" context includes thread name, for some client IP, for another it's the dialect or maybe SQL hints or a `ThreadLocal` `ReplicaConsistency` override or a bajillion different things. These things will be attracted to this class.

But we don't need all of that. What we *use* this for is to know *why* `db-replica` decided to route to main or replica. So maybe:
```suggestion
public final class RouteDecision {
```

src/main/java/com/atlassian/db/replica/api/context/Reason.java
```diff
@@ -0,0 +1,43 @@
...
+public final class Reason {
+    private final String value;
+
+    private Reason(final String value) {
+        this.value = value;
+    }
+
+    public static final Reason API_CALL = new Reason("API_CALL");
+    public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT");
+    public static final Reason READ_OPERATION = new Reason("READ_OPERATION");
```
dagguh, line 14: It seems that this is silently coupled with `QueryContext.runOnMain = false`.
In other words, if we see a `QueryContext(READ_OPERATION, true)` then it's a bug. Right?
If so, let's make this coupling explicit, keep it local, e.g. by adding the boolean here.

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
wyrzyk: `SELECT ... FOR UPDATE` is [a form of lock](https://www.postgresql.org/docs/9.4/explicit-locking.html). I don't think we need a better granularity. There's SQL in the context to determine what kind of lock causes it.

wyrzyk:
@dagguh Thanks for the feedback. Lots of good ideas!

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -80,7 +82,7 @@ public boolean getReadOnly() {
     public void setReadOnly(boolean readOnly) throws SQLException {
         isReadOnly = readOnly;
         if (readOnly) {
-            state.getReadConnection().setReadOnly(isReadOnly);
+            state.getReadConnection(QueryContext.builder(Reason.API_CALL,false)).setReadOnly(isReadOnly);
```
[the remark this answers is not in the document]
dagguh: Actually, if we add a "cause" to `API_CALL`, then we can reuse `API_CALL`

src/main/java/com/atlassian/db/replica/api/context/Reason.java
wyrzyk: - [ ] keep it local, e.g. by adding the boolean here.

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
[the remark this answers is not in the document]
wyrzyk: Any idea why would we care about this level of details? Any example of a use case?

src/main/java/com/atlassian/db/replica/internal/ReplicaPreparedStatement.java
```diff
@@ -78,22 +67,33 @@ protected ReplicaPreparedStatement(
     @Override
     public ResultSet executeQuery() throws SQLException {
         checkClosed();
-        final PreparedStatement statement = getReadStatement(sql);
-        return execute(statement::executeQuery);
+        final QueryContext.Builder contextBuilder = QueryContext.builder(Reason.READ_OPERATION, false).sql(sql);
```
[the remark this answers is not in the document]
wyrzyk: > maybe getReadStatement could hold that default?

Yes. I think getReadStatement could hold the default.

> If not, then at least it could read the sql from context, because now it gets it via both params.

Context#Builder is an API, and the SQL was needed only for internal usage here.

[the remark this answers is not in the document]
wyrzyk: Maybe Context#Builder  shouldn't be part of API 🤔

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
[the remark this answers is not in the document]
dagguh: We could see we use main due to too many `executeUpdate` calls or too many `executeBatch`. We could then focus on usages of that specific method in Jira. There are a lot of signatures of methods with those names, so this amount of detail could indeed be unnecessary. If we assume that these calls are unchangeable, then indeed the drilldown doesn't help.

src/main/java/com/atlassian/db/replica/api/context/Reason.java
dagguh: If we do that then we have to split `API_CALL` to one for `runOnMain = true` and another for `false`

src/main/java/com/atlassian/db/replica/internal/ReplicaPreparedStatement.java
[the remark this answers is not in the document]
dagguh: if we don't have a concrete case for new API, we should hide it

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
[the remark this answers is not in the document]
wyrzyk: IMO we need to distinguish between RO_API calls and MAIN_API calls, for now, we can add more details later if needed (and I don't think it will be needed if we mix it with stack-trace mechanism).

src/main/java/com/atlassian/db/replica/api/context/Reason.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static final Reason API_CALL = new Reason("API_CALL", true);
+    public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT", true);
+    public static final Reason READ_OPERATION = new Reason("READ_OPERATION", false);
+    public static final Reason WRITE_OPERATION = new Reason("WRITE_OPERATION", true);
+    public static final Reason LOCK = new Reason("LOCK", true);
+    public static final Reason MAIN_CONNECTION_REUSE = new Reason(
+        "MAIN_CONNECTION_REUSE",
+        true
+    ); // TODO: Add cause to this reason!! (an sql/reason that is responsible for the switch)
```
dagguh, line 25: - [ ] TODO: Add cause to this reason!! (an sql/reason that is responsible for the switch)

src/main/java/com/atlassian/db/replica/internal/ForwardCall.java
```diff
@@ -1,18 +1,16 @@
...
-public class ForwardCall implements DualCall {
-    @Override
-    public <T> T callReplica(SqlCall<T> call) throws SQLException {
-        return call.call();
-    }
+public class ForwardCall implements DatabaseCall {
 
     @Override
-    public <T> T callMain(SqlCall<T> call) throws SQLException {
+    public <T> T call(final SqlCall<T> call, RouteDecision context) throws SQLException {
```
dagguh, line 12:
```suggestion
    public <T> T call(final SqlCall<T> call, RouteDecision decision) throws SQLException {
```

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -37,8 +38,8 @@ public Connection getWriteConnection() throws SQLException {
         return state.getWriteConnection();
     }
 
-    public Connection getReadConnection() throws SQLException {
-        return state.getReadConnection();
+    public Connection getReadConnection(RouteDecisionBuilder contextBuilder) throws SQLException {
```
dagguh, line 41:
```suggestion
    public Connection getReadConnection(RouteDecisionBuilder decisionBuilder) throws SQLException {
```

src/main/java/com/atlassian/db/replica/internal/ReplicaPreparedStatement.java
```diff
@@ -78,22 +81,23 @@ protected ReplicaPreparedStatement(
...
+        final PreparedStatement statement = getReadStatement(contextBuilder);
+        return execute(statement::executeQuery, contextBuilder.build());
     }
 
     @Override
     public int executeUpdate() throws SQLException {
         checkClosed();
         final PreparedStatement statement = getWriteStatement();
-        return execute(statement::executeUpdate);
+        return execute(statement::executeUpdate, new RouteDecisionBuilder(API_CALL).sql(sql).build());
```
dagguh, line 93: Rename to `RW_API_CALL` to distinguish from `RO_API_CALL`?

src/main/java/com/atlassian/db/replica/internal/ReplicaPreparedStatement.java
```diff
@@ -544,8 +548,8 @@ public PreparedStatement getWriteStatement() {
     }
 
     @Override
-    public PreparedStatement getReadStatement(String sql) {
-        return (PreparedStatement) super.getReadStatement(sql);
+    public PreparedStatement getReadStatement(RouteDecisionBuilder contextBuilder) {
+        return (PreparedStatement) super.getReadStatement(contextBuilder);
```
dagguh, line 552:
```suggestion
    public PreparedStatement getReadStatement(RouteDecisionBuilder decisionBuilder) {
        return (PreparedStatement) super.getReadStatement(decisionBuilder);
```

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -426,47 +438,58 @@ public long getLargeMaxRows() throws SQLException {
...
-            return dualCall.callReplica(call);
-        } else {
-            final T result = dualCall.callMain(call);
-            if (isWriteOperation) {
-                recordWriteAfterQueryExecution();
-            }
-            return result;
+    <T> T execute(final SqlCall<T> call, final RouteDecision routeDecision) throws SQLException {
+        final T result = databaseCall.call(call, routeDecision);
+        if (routeDecision.getReason().isRunOnMain() && isWriteOperation) {
```
dagguh, line 489: Nice

src/main/java/com/atlassian/db/replica/internal/RouteDecisionBuilder.java
```diff
@@ -0,0 +1,56 @@
+package com.atlassian.db.replica.internal;
+
+import com.atlassian.db.replica.api.context.Reason;
+import com.atlassian.db.replica.api.context.RouteDecision;
+
+import java.util.Objects;
+
+public final class RouteDecisionBuilder {
+    private Reason reason;
+    private String sql;
```
dagguh, line 10: This can be `null`. Go for `Optional<String>` instead?

src/main/java/com/atlassian/db/replica/internal/RouteDecisionBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+    public RouteDecisionBuilder sql(final String sql) {
+        this.sql = sql;
+        return this;
+    }
+
+    public String getSql() {
+        return sql;
+    }
+
+    public RouteDecision build() {
```
dagguh, line 30: This builder doesn't provide defaults. It's used to override the `reason` or optionally attach an `sql`.
We could reduce boilerplate by removing the builder and adding `RouteDecision withReason(Reason reason)` to `RouteDecision` itself. Same for `withSql(String sql)`.

Unless we need to distinguish "partial" decision from a full decision explicitly.

src/main/java/com/atlassian/db/replica/spi/DatabaseCall.java
```diff
@@ -0,0 +1,13 @@
...
+import com.atlassian.db.replica.api.SqlCall;
+import com.atlassian.db.replica.api.context.RouteDecision;
+
+import java.sql.SQLException;
+
+/**
+ * Intercepts call to a database.
+ */
+public interface DatabaseCall {
+    <T> T call(final SqlCall<T> call, RouteDecision context) throws SQLException;
```
dagguh, line 12:
```suggestion
    <T> T call(final SqlCall<T> call, RouteDecision decision) throws SQLException;
```

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -156,11 +160,13 @@ public void close() throws SQLException {
     /**
      * Provides a connection that will be used for reading operation. Will use read-replica if possible.
      */
-    private Connection prepareReadConnection() throws SQLException {
+    private Connection prepareReadConnection(RouteDecisionBuilder contextBuilder) throws SQLException {
```
dagguh, line 163:
```suggestion
    private Connection prepareReadConnection(RouteDecisionBuilder decisionBuilder) throws SQLException {
```

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -92,9 +96,9 @@ public State getState() {
     /**
      * Provides a connection that will be used for reading operation. Will use read-replica if possible.
      */
-    public Connection getReadConnection() throws SQLException {
+    public Connection getReadConnection(RouteDecisionBuilder contextBuilder) throws SQLException {
```
dagguh, line 99:
```suggestion
    public Connection getReadConnection(RouteDecisionBuilder decisionBuilder) throws SQLException {
```

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -603,58 +642,81 @@ public void shouldHoldOnlyOneConnection() throws SQLException {
...
 
         connection.prepareStatement(SIMPLE_QUERY).executeQuery();
 
         assertThat(connectionProvider.getProvidedConnectionTypes())
             .containsOnly(REPLICA, MAIN);
-        verify(dualCall).callMain(any());
-        verify(dualCall, never()).callReplica(any());
+        verify(databaseCall).call(
+            any(),
+            eq(new RouteDecisionBuilder(Reason.REPLICA_INCONSISTENT).sql(SIMPLE_QUERY).build())
```
dagguh, line 679: Looks good!

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -513,14 +536,24 @@ void recordWriteAfterQueryExecution() throws SQLException {
         }
     }
 
-    public Statement getReadStatement(String sql) {
+    public Statement getReadStatement(RouteDecisionBuilder contextBuilder) {
```
dagguh, line 539:
```suggestion
    public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) {
```

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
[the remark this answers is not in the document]
dagguh: Agreed

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: Thanks

src/main/java/com/atlassian/db/replica/api/context/Reason.java
wyrzyk: There's already a task for it

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -603,58 +642,88 @@ public void shouldHoldOnlyOneConnection() throws SQLException {
...
+            connectionProvider,
+            permanentConsistency().build()
+        ).databaseCall(databaseCall)
+            .build();
+        dualConnection.prepareStatement(SIMPLE_QUERY).executeUpdate();
+        Mockito.reset(databaseCall);
+
+        dualConnection.prepareStatement(SIMPLE_QUERY).executeQuery();
+
+        verify(databaseCall).call(
```
wyrzyk, line 715: @dagguh Look at this!

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -603,58 +642,88 @@ public void shouldHoldOnlyOneConnection() throws SQLException {
...
+            any(),
+            eq(
+                new RouteDecisionBuilder(Reason.MAIN_CONNECTION_REUSE)
+                    .sql(SIMPLE_QUERY)
+                    .currentStateDecision(
+                        new RouteDecisionBuilder(Reason.API_CALL).sql(SIMPLE_QUERY).build()
```
wyrzyk, line 721: I think about adding more info to API_CALL reason 🤔

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
dagguh: Oooooh ✨ 
So it's like a reason chain just like exception cause chain. Very nice!

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public void reset() {
+        this.lazyReference.reset();
+        this.decisionBuilder.set(null);
+    }
+
+    public RouteDecisionBuilder getDecisionBuilder() {
+        return decisionBuilder.get();
+    }
```
dagguh, line 29: So now we're tracking `RouteDecision.currentStateDecision`, so we have state, so we have to reset it so we have this reference class for that 👍

src/main/java/com/atlassian/db/replica/internal/RouteDecisionBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.atlassian.db.replica.api.context.Reason;
+import com.atlassian.db.replica.api.context.RouteDecision;
+
+import java.util.Objects;
+
+public final class RouteDecisionBuilder {
+    private String sql;
+    private Reason reason;
+    private RouteDecision currentStateDecision;
```
dagguh, line 11: As a fan of short names, I gotta try:
```suggestion
    private RouteDecision cause;
```
😉

dagguh, line 11: Could be null, so:
```suggestion
    private Optional<RouteDecision> cause;
```

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
```diff
@@ -0,0 +1,44 @@
...
+        return lazyReference.get();
+    }
+
+    public void reset() {
+        this.lazyReference.reset();
+        this.decisionBuilder.set(null);
+    }
+
+    public RouteDecisionBuilder getDecisionBuilder() {
+        return decisionBuilder.get();
```
dagguh, line 28: Returns null if `!isInitialized`. Go `Optional`?

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
```diff
@@ -0,0 +1,44 @@
+package com.atlassian.db.replica.internal;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class DecisionAwareReference<T> {
+    private final LazyReference<T> lazyReference = new InnerLazyReference<>(this);
```
dagguh, line 6: I'm having trouble understanding why `LazyReference` wouldn't work

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
```diff
@@ -0,0 +1,44 @@
...
+    private final AtomicReference<RouteDecisionBuilder> decisionBuilder = new AtomicReference<>();
+
+    public abstract T create() throws Exception;
```
dagguh, line 9:
```suggestion
    protected abstract T create() throws Exception;
```

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
wyrzyk: Can you try to refactor? Maybe it's possible to use LazyReference, and I didn't figure out how.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -513,15 +555,25 @@ void recordWriteAfterQueryExecution() throws SQLException {
         }
     }
 
-    public Statement getReadStatement(String sql) {
+    public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) {
+        if (connectionProvider.getState().equals(MAIN)) {
+            decisionBuilder.reason(MAIN_CONNECTION_REUSE);
+            connectionProvider.getStateDecision().ifPresent(decisionBuilder::cause);
```
dagguh, line 561: New reason for deleting builder: it's a mutable object being passed around. With just an immutable `RouteDecision`, we'd have:
```suggestion
            RouteDecision chainedDecision = connectionProvider.getStateDecision().map(decision::withCause).orElse(decision);
```

wyrzyk: What do you mean by deleting the builder? The main reason for the builder is not to reveal unnecessary API like witchCause with the RouteDecission. There's no use case for this API for consumers of the library.

wyrzyk: We can make the builder immutable if it's the concern.

dagguh: > The main reason for the builder is not to reveal unnecessary API like witchCause with the RouteDecission.

I didn't realize that. Thanks

dagguh: Exposing `withCause` wouldn't hurt us, because the object is still immutable and it doesn't leak any new data. The lib consumers can reimplement that trivially. Hiding it doesn't help maintainability that much. Without the builder, the source of the impl could be easier to read.
It's a preference thing, not a must-have, not required for the merge.

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
dagguh: So now we have:
```java
if (this.decisionBuilder.get() == null) {
    throw new IllegalStateException("The decision builder is not initialized");
}
return decisionBuilder.get();
```
So still a runtime error. I suggested a compile-time error. I'll include it in my PR. If we want this to be a runtime error, let me know.

dagguh: Hmm, ok, there's no sensible handling of an empty (missing) decision. So right now it relies on `get` being called earlier (which is safe to assume within `create`). I'll think if I can simplify this assumption

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class DecisionAwareReference<T> extends LazyReference<T> {
+    private final AtomicReference<RouteDecisionBuilder> decisionBuilder = new AtomicReference<>();
+
+    public T get(RouteDecisionBuilder decisionBuilder) {
+        if (!isInitialized()) {
+            this.decisionBuilder.compareAndSet(null, decisionBuilder);
+        }
+        return super.get();
```
dagguh, line 12: So when `isInitialized()`, then we ignore the `decisionBuilder`. Intentional?

wyrzyk: We need to store the cause when the connection is initialised. Otherwise, it will not be a cause. OTOH we can ignore it as CAS will change the value only once.

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
wyrzyk: I made 2 attempts to make it a compile-time error. It led to a broad refactoring. IMO the PR is already complex, and the internals can be refactored later.

wyrzyk:
@dagguh The conversation is broad, and some of the comments are already outdated. What blocks this PR from merging?

src/main/java/com/atlassian/db/replica/internal/ForwardCall.java
wyrzyk: It's outdated. I'm resolving the suggestion.

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
dagguh: Ok, simplified a little: https://github.com/wyrzyk/db-replica/pull/4

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
dagguh: Fair enough. It's internal anyway 👍

dagguh:
What's left:
* finished context rename: https://github.com/wyrzyk/db-replica/pull/5
* consider removing the builder (PR incoming)

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: The default decision could bring `sql`, so it cannot be inlined here. Gotta let the `getReadStatement(RouteDecisionBuilder decisionBuilder)` parameter be

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: https://github.com/wyrzyk/db-replica/pull/6

dagguh (approved):
https://github.com/wyrzyk/db-replica/pull/6 causes behavioral problems. They can be fixed, but it's non-trivial effort. I might come back to that later.

Issue #23, opened 2021-02-16
Some read-only queries go to the main database instead of replica 


wyrzyk:
Example:
`SELECT count(*) FROM user`

PR #24, opened 2021-02-16
#23: Support read-only functions

wyrzyk:


docs/switching-between-main-and-replica.md
```diff
@@ -15,5 +15,7 @@ will switch the connection's state to the main database.
     - `SELECT FOR UPDATE` statement.
     - `UPDATE` statement.
     - `DELETE` statement.
-    - an SQL function call.
     - All calls with transaction isolation level higher than `TRANSACTION_READ_COMMITTED`.
+
+6.  The query will use the main database in case it's an unknown function call. Known read-only functions are [standard
```
wyrzyk, line 20: I need to obtain the list of standard SQL read-only functions. The most up to date SQL standard I have found is SQL-92. The newest standards seem to be protected by a paywall.

wyrzyk: - [ ] Obtain the list of standard SQL read-only functions

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -19,25 +13,8 @@
...
-import java.sql.SQLException;
-import java.sql.SQLWarning;
-import java.sql.SQLXML;
-import java.sql.Savepoint;
-import java.sql.Statement;
-import java.sql.Struct;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
+import java.sql.*;
```
wyrzyk, line 16: Should I update my settings?

docs/switching-between-main-and-replica.md
wyrzyk: I think I'll start with https://www.techonthenet.com/postgresql/functions/index.php. We can add more later.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -251,6 +252,38 @@ public void shouldUseMainConnectionForSelectFunctionCalls() throws SQLException
...
 
+    @Test
+    public void shouldUseReplicaForKnownReadOnlyFunctionCalls() throws SQLException {
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).build();
+
+        connection.prepareStatement("SELECT count(*) FROM user").executeQuery();
```
dwatl, line 263: Could we use the queries that caused the problem in the first place?

dwatl: Red-green-refactor would be nice though

wyrzyk: > Could we use the queries that caused the problem in the first place?

I don't want to use in-product queries in OS lib. I don't have time and energy to think if I'm not leaking too much.

> Red-green-refactor would be nice though

It's nice but also has drawbacks. With RGR master is not green all the. It makes git bisect harder. You can always rollback non-test changes and do RGR while reviewing the change.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dagguh: That or wait for [the fix](https://youtrack.jetbrains.com/issue/IDEA-225733#focus=Comments-27-4627146.0-0) in IJ 2021.1

docs/switching-between-main-and-replica.md
dagguh: Good call

dagguh (approved):

dagguh:
Red CI:
```
java.lang.AssertionError: [thruput per ms] 
Expecting:
 <2766.3284f>
to be greater than:
 <3000.0f>
```

Cost of the feature?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -50,34 +27,49 @@
...
     @Override
     public CallableStatement prepareCall(String sql) throws SQLException {
         checkClosed();
-        return new ReplicaCallableStatement.Builder(connectionProvider, consistency, databaseCall, sql).build();
+        return new ReplicaCallableStatement.Builder(
+            connectionProvider,
+            consistency,
+            databaseCall,
+            sql,
+            readOnlyFunctions
```
dagguh, line 71: This should be optional (empty list is a good default), so it shouldn't be required by the builder constructor

wyrzyk:
> Cost of the feature?

I'm not sure. It was flaky recently. I'm not satisfied with the stability of the test. Maybe we can utilise JPT to test the performance of the library. WDYT?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -600,6 +663,9 @@ private boolean isFunctionCall(String sql) {
             return false;
         }
         final String potentialMethodName = mayContainFunction.split(methodBracketStart)[0];
+        if (standardReadOnlyFunctions.contains(potentialMethodName) || readOnlyFunctions.contains(potentialMethodName)) {
```
dagguh, line 666: `potentialMethodName` sounds fuzzy, but the `readOnlyFunctions` is a strict `contains`. Does this actually catch function names in a more messy query?

- [ ] test a case where first function is RO, but second is RW

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/api/DualConnection.java
wyrzyk: I want to always propagate `readOnlyFunctions` from DualConnectin to Statements. Not propagating it is a bug. I don't want it to have its own default. It's not API. It's internal.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
wyrzyk: I know it's technically possible to have cases like this, but please find one from the production code. I don't think safety for the cases that may never occur is worth the extra complexity.

>  test with lowercase and uppercase function names

Lower and upper case functions are supported for the built-in functions, and it's covered. Supporting it for user-defined functions is up to the library user. If someone uses both upper and lower case invocations of the same method in the codebase, then both ways need to be registered.

dagguh:
> Maybe we can utilise JPT to test the performance of the library. WDYT?

Testing on Jira? That could work. Synthetic load is never perfect, but it's infinitely more convincing than a microbenchmark. So either that or A/B tests on staging.

wyrzyk:
> Testing on Jira? 

I mean setting up infrastructure with JPT, not testing with a product.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: 👍 case sensitivity is indeed covered:
```java
        connection.prepareStatement("SELECT max(user.id) FROM user").executeQuery();
        connection.prepareStatement("SELECT MAX(user.id) FROM user").executeQuery();
```

dagguh: > please find one from the production code

I cannot find it. Perhaps sanitizing it and putting it in the test would not be such a bad idea.

> I don't think safety for the cases that may never occur is worth the extra complexity.

How do we know these cases do not occur right now? Is a mix of RO and RW functions unrealistic? If so, why?

dagguh:
I see, so a DB server on EC2 and a thin JDBC client consuming `DualConnection`? We can do that. We'd need to:
* populate the DB with data (maybe there's an opensource dataset?)
* come up with SQL queries to run

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
wyrzyk: > How do we know these cases do not occur right now? Is a mix of RO and RW functions unrealistic? If so, why?

We will know because we'll see an exception if we try to run RW operation on the replica. If it's realistic, please find one case like this in the logs we have.

dagguh: We see lots of exceptions though, tens, hundreds, thousands, depending on the day.

dagguh: But ok, one step at a time.
If these exceptions are due to insufficient function detection, then this PR does not make that worse.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -39,6 +30,408 @@
     private final DatabaseCall databaseCall;
     final String methodBracketStart = Pattern.quote("(");
     private boolean isWriteOperation = true;
+    private final Set<String> readOnlyFunctions;
+    private static final Set<String> standardReadOnlyFunctions = Stream
```
dagguh, line 34: Ok, this guy is big
We can extract an `SqlFunctions` class, which could hold the list and the query parsing logic. Should be easier to unit-test too

wyrzyk: I was waiting for this comment 😉 . But seriously I'm testing it with the product, and I postponed refactoring.

wyrzyk:
>  Verify if it solves the original issue

It turned out that it's not enough to solve the original issue. I needed to apply more fixes
(https://github.com/atlassian-labs/db-replica/pull/26, https://github.com/atlassian-labs/db-replica/pull/25). Now when I test the read-only page it goes 100% to the replica.

dagguh:
PS. You can put the issue number in the commit message like:

```
Add support for read-only functions

Fix #23.
```

dagguh:
The [merge commit](https://github.com/atlassian-labs/db-replica/pull/24/commits/d8d6fed556312e3dda643ebd0776956b9858425a) is a result of resolving the conflict in GH GUI. Feel free to squash-merge this or force-push a rebase.

dagguh (dismissed):

dagguh:
Hmm, new conflicts :)

wyrzyk:
> Hmm, new conflicts :)

resolved

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -597,28 +587,6 @@ private boolean isDelete(String sql) {
...
-    }
-
-    /**
-     * Skips `SELECT ` at the beginning of the query. Postgres identifiers are limited to
-     * 63 characters, so we should be safe to interpret first 80 characters.
-     */
-    private String skipIrrelevantSqlParts(String sql) {
-        return sql.substring(7, min(sql.length(), 80));
-    }
-
```
dagguh, line 621: ❤️

tciesla-atlassian (approved):
LGTM

src/main/java/com/atlassian/db/replica/internal/SqlFunction.java
```diff
@@ -0,0 +1,443 @@
...
+        )
+        .flatMap(name -> Stream.of(name, name.toUpperCase()))
+        .collect(Collectors.toCollection(HashSet::new));
+    private final Set<String> readOnlyFunctions;
+
+    public SqlFunction(Set<String> readOnlyFunctions) {
+        this.readOnlyFunctions = readOnlyFunctions;
+    }
+
+    boolean isFunctionCall(String sql) {
```
dagguh, line 418: Returns `false` for an RO function call.

WDYT?
```java
static Optional<SqlFunction> getFirstFunctionCall(String sql) {}
boolean isWritingToDatabase() {}
```

dagguh (approved):

src/main/java/com/atlassian/db/replica/internal/SqlFunction.java
```diff
@@ -0,0 +1,443 @@
...
+            "to_regoperator",
+            "to_regtype",
+            "col_description",
+            "shobj_description",
+            "txid_current",
+            "txid_current_snapshot",
+            "txid_snapshot_xip",
+            "txid_snapshot_xmax",
+            "txid_snapshot_xmin",
+            "txid_visible_in_snapshot"
```
afaruga-atlassian, line 408: @wyrzyk How we would discover the missing function here? other than manual 1:1 analysis, once someone discovers that sth should be faster.
As a systematic way, I'd imagine getting a list of unique calls done by Jira within 1 workweek day and weekend and filter functions to find out potential missed.

Though when I write about it I already started thinking that this is over-engineering.

dagguh: We could emit a `new RouteDecision(Reason.FUNCTION_CALL, functionName)`. We'd be abusing the `String sql` field, but it could be generalized.

wyrzyk: @afaruga-atlassian, @dagguh Thanks for the feedback. The diagnostics API is still in progress. We currently see we miss API method name for `Reason.RW_API_CALL`, and functionName for `Reason.FUNCTION_CALL`. I think about adding it to be more generic (3rd use case would help here).

dagguh: 3rd use case: SQL for `Reason.WRITE_OPERATION` (due to `isUpdate(sql) || isDelete(sql)`).
Do we need the SQL for other `Reason`s?

PR #26, opened 2021-02-18
Avoid unnecessary switch to `MainConnection` while setting `read-only` mode

wyrzyk:
As per JavaDoc `read-only` mode is just a hint to the driver to enable
optimisations. In the case of `DualConnection` switching to the main just in case
is anti-optimisation.

I found out we set it to `false` just in case in `DatabaseAccessorImpl#runInTransaction`.
There might be more places like this, that prevent us from utilising replica more.

### DIFF
--- a/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
+++ b/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
@@ -783,9 +783,24 @@ public class TestDualConnection {
         connection.setReadOnly(false);
 
         assertThat(connectionProvider.getProvidedConnectionTypes())
-            .containsOnly(MAIN);
+            .containsOnly(REPLICA);
         verify(connectionProvider.singleProvidedConnection()).setReadOnly(false);
         assertThat(connection.isReadOnly()).isFalse();
+    }
+
+    @Test
+    public void shouldUtiliseReplicaEvenAfterDisablingReadOnly() throws SQLException {
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).build();
+
+        connection.setReadOnly(false);
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+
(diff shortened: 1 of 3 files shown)


src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -783,25 +783,39 @@ public void shouldStopReadOnlyMode() throws SQLException {
...
         assertThat(connectionProvider.getProvidedConnectionTypes())
-            .containsOnly(MAIN);
+            .containsOnly(REPLICA);
         verify(connectionProvider.singleProvidedConnection()).setReadOnly(false);
         assertThat(connection.isReadOnly()).isFalse();
     }
 
     @Test
-    public void shouldNotSwitchBackToReplica() throws SQLException {
+    public void shouldUtiliseReplicaEvenAfterDisablingReadOnly() throws SQLException {
```
dagguh, line 792: Sweet

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -783,25 +783,39 @@ public void shouldStopReadOnlyMode() throws SQLException {
...
         connection.setReadOnly(false);
-        connection.setReadOnly(true);
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+
+        assertThat(connectionProvider.getProvidedConnectionTypes())
+            .containsOnly(REPLICA);
+    }
+
+    @Test
+    public void shouldNotSwitchBackToReplica() throws SQLException {
```
dagguh, line 807: This test case seems unrelated to the PR. Correct?

dagguh (approved):

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: Yes. The test case is unrelated. The diff doesn't show it, but   `shouldNotSwitchBackToReplica` was reusing `setReadOnly` to switch between the main database and replica. It will not work after the fix, so I changed it to use `executeUpdate` and `executeQuery`.

PR #30, opened 2021-02-22
Set connection properties while switching to replica

wyrzyk:
Fix for #29

- [x] Rewrite it as a unit test
- [x] Fix the issue
- [x] Update the changelog


wyrzyk:
>  Rewrite it as a unit test

I'll leave it out of the scope for this PI. I'll fix it in the next one.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -184,6 +184,7 @@ private Connection prepareReadConnection(RouteDecisionBuilder decisionBuilder) t
         if (consistency.isConsistent(() -> readConnection.get(decisionBuilder))) {
             if (getState().equals(COMMITED_MAIN)) {
                 closeConnection(writeConnection, decisionBuilder);
+                parameters.initialize(readConnection.get(decisionBuilder));
```
dagguh, line 187: Removing the fix is detected by the new test 👍 
`org.postgresql.util.PSQLException: Cannot commit when autoCommit is enabled.`

dagguh: When I bring it out of the `if`s, it's green.

- [ ] move outside of ifs or catch in tests

dagguh (changes requested):

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -247,7 +247,6 @@ public void shouldUseReplicaForKnownReadOnlyFunctionCalls() throws SQLException
         connection.prepareStatement("SELECT COUNT(*) FROM user").executeQuery();
         connection.prepareStatement("SELECT max(user.id) FROM user").executeQuery();
         connection.prepareStatement("SELECT MAX(user.id) FROM user").executeQuery();
-        connection.prepareStatement("SELECT nextval('foo')").executeQuery();
```
dagguh, line 250: It makes sense. I didn't catch this earlier 😬

src/main/java/com/atlassian/db/replica/internal/SqlFunction.java
```diff
@@ -238,9 +238,6 @@
             "jsonb_to_record",
             "json_to_recordset",
             "jsonb_to_recordset",
-            "currval",
-            "lastval",
-            "nextval",
```
dagguh, line 243: Reverting the fix is caught by tests 👍 
`org.postgresql.util.PSQLException: ERROR: cannot execute nextval() in a read-only transaction`

dagguh (approved):

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: It makes sense to populate the parameters for every change to the replica.

dagguh: That was the last place where `State.NOT_INITIALISED` was explicitly checked. But we should preserve it, because it's still implicitly checked in `else` clauses like in `if (state.equals(REPLICA)) else { ... }`. And it's still exposed as API.

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
```diff
@@ -20,4 +25,63 @@ public void shouldUseReplica() throws SQLException {
...
+        Connection mainConnection,
+        LsnReplicaConsistency consistency
+    ) {
+        consistency.write(mainConnection);
+        for (int i = 0; i < 30; i++) {
+            if (consistency.isConsistent(connectionProvider::getReplicaConnection)) {
+                return;
+            }
+            try {
+                Thread.sleep(1000);
```
afaruga-atlassian, line 79: Would it generate some flakiness in the future? It's a good trade-off for now, though it's nto sth like `CountDownLatch`

dagguh: You mean spurious wake-ups?

dagguh: This is a simple impl of a timeout. As an integration test, a real `Clock.systemUTC()` is used, so we need time to pass. It's not like we can inject a latch into various participants and make them cooperate

wyrzyk: I wrote an integration test to be sure I'm solving the same issue we see on prod. I plan to rewrite it as a unit test.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: @dagguh I see we have the same issue with the write connection. It seems it can pass without initialisation. I'm working on the repro.

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
afaruga-atlassian: Fair enough, below convinced me, thanks for explaining the reasoning behind trade-off. 👍 
> As an integration test, a real Clock.systemUTC() is used

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
dagguh: aye aye

Issue #32, opened 2021-02-23
org.postgresql.util.PSQLException: ERROR: cannot execute UPDATE in a read-only transaction


wyrzyk:

wyrzyk:
Fixed in `0.1.27`.

PR #34, opened 2021-02-23
Propagate the read-only mode to the main connection

wyrzyk:
Fix for #32

### DIFF
--- a/src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
@@ -76,7 +76,7 @@ public final class ConnectionParameters {
         this.holdability = holdability;
     }
 
-    public Boolean isReadOnly() {
+    public boolean isReadOnly() {
         return readOnly != null && readOnly;
     }
 
--- a/src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
@@ -9,6 +9,7 @@ import java.util.Map;
 public final class ConnectionParameters {
 
     private Boolean isAutoCommit;
+    private Boolean readOnly;
     private Integer transactionIsolation;
     private String catalog;
     private Map<String, Class<?>> typeMap;
@@ -29,6 +30,9 @@ public final class ConnectionParameters {
         }
         if (holdability != null) {
             connection.setHoldability(holdability);
+        }
+        if (readOnly != null) {
+            connection.setReadOnly(readOnly);
         }
     }
 
@@ -72,10 +76,19 @@ public final class ConnectionParameters {
         this.holdability = holdability;
     }
 
+    public Boolean isReadOnly() {
+        return readOnly != null && readOnly;
+    }
+
+    public void setReadOnly(boolean readOnly) {
+        this.readOnly = readOnly;
+    }
+
     @Override
     public String toString() {
         return "ConnectionParameters{" +
             "isAutoCommit=" + isAutoCommit +
+            ", isReadOnly=" + readOnly +
             ", transactionIsolation=" + transactionIsolation +
             ", catalog='" + catalog + '\'' +
             ", typeMap=" + typeMap +
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
@@ -22,7 +22,6 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     private final ReplicaConsistency consistency;
     private final ConnectionState state;
     private final ConnectionParameters parameters;
-    private Boolean isReadOnly;
     private final Warnings warnings;
 
     public ReplicaConnectionProvider(
@@ -77,12 +76,11 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     }
 
     public boolean getReadOnly() {
-        return isReadOnly != null && isReadOnly;
+        return parameters.isReadOnly();
     }
 
     public void setReadOnly(boolean readOnly) throws SQLException {
-        this.isReadOnly = readOnly;
-        state.getReadConnection(new RouteDecisionBuilder(RO_API_CALL)).setReadOnly(readOnly);
+        parameters.setReadOnly(readOnly);
     }
 
     public String getCatalog() {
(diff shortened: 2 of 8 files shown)


src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -72,10 +76,19 @@ public void setHoldability(Integer holdability) {
         this.holdability = holdability;
     }
 
+    public boolean isReadOnly() {
+        return readOnly != null && readOnly;
```
dagguh, line 80: So `false` by default, but we cannot inline that in the field initializer, because we need the `null` in `initialize`

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -77,12 +76,11 @@ public Boolean isClosed() {
     }
 
     public boolean getReadOnly() {
-        return isReadOnly != null && isReadOnly;
+        return parameters.isReadOnly();
     }
 
     public void setReadOnly(boolean readOnly) throws SQLException {
-        this.isReadOnly = readOnly;
-        state.getReadConnection(new RouteDecisionBuilder(RO_API_CALL)).setReadOnly(readOnly);
```
dagguh, line 85: Oh, so we don't send it to the connection now, I see

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -131,11 +131,10 @@ private Connection prepareMainConnection(RouteDecisionBuilder decisionBuilder) t
         final Optional<Connection> connection = getConnection();
         if (connection.isPresent() && connection.get().equals(writeConnection.get(decisionBuilder))) {
             readConnection.reset();
-            parameters.initialize(writeConnection.get(decisionBuilder));
         } else {
             closeConnection(readConnection, decisionBuilder);
-            parameters.initialize(writeConnection.get(decisionBuilder));
         }
+        parameters.initialize(writeConnection.get(decisionBuilder));
```
dagguh, line 137: This is So this is the "escape the ifs" counterpart for `writeConnection` 👍

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1382,4 +1401,21 @@ public void shouldKeepUsingReplicaAfterSettingTimeout() throws SQLException {
...
+            permanentConsistency().build()
+        ).build();
+        connection.setReadOnly(true);
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+        final Connection rawConnection = connectionProvider.singleProvidedConnection();
+        Mockito.reset(rawConnection);
+
+        connection.close();
+
+        Mockito.verify(rawConnection).setReadOnly(eq(false));
```
dagguh, line 1419: How can a closed connection state affect anything? Can it be reopened?

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
```diff
@@ -43,45 +47,48 @@ public void shouldPreserveAutoCommitModeWhileSwitchingFromReplicaToMain() throws
     }
 
     @Test
-    public void shouldRunNextValOnMainDatabase() throws SQLException {
```
dagguh, line 46: This mixes up the diff a little. Can we avoid unnecessary "minus lines" in the diff?

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
```diff
@@ -43,45 +47,48 @@ public void shouldPreserveAutoCommitModeWhileSwitchingFromReplicaToMain() throws
...
-        try (final Statement mainStatement = mainConnection.createStatement()) {
+    private void createTestSequence(Connection connection) throws SQLException {
+        try (final Statement mainStatement = connection.createStatement()) {
             mainStatement.execute("CREATE SEQUENCE test_sequence;");
-            final LsnReplicaConsistency consistency = new LsnReplicaConsistency();
-            waitForReplicaConsistency(connectionProvider, mainConnection, consistency);
         }
     }
 
-    private void waitForReplicaConsistency(
```
dagguh, line 68: Extracted to `WaitingReplicaConsistency ` 👍

dagguh (approved):

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1382,4 +1401,21 @@ public void shouldKeepUsingReplicaAfterSettingTimeout() throws SQLException {
...
+    public void shouldResetReadOnlyModeBeforeReleasingConnection() throws SQLException {
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).build();
+        connection.setReadOnly(true);
+        connection.prepareStatement(SIMPLE_QUERY).executeQuery();
+        final Connection rawConnection = connectionProvider.singleProvidedConnection();
+        Mockito.reset(rawConnection);
```
afaruga-atlassian, line 1415: why do you have to reset mock ?

afaruga-atlassian (approved):

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
dagguh: Reset focuses on the `BeforeReleasingConnection` part of the test case

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: We are not the only ones adding magic between `jdbc` and business logic 😄 . connection pool intercepts connection#close and re-uses it instead of closing.

dagguh: Aaah 🤯 
Good point! Mental model updated 😉

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
wyrzyk: Maybe not exactly this implementation, but something similar could work in production code for some not critical, but SQL heavy code paths (generating reports, backups ...).

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
dagguh: I'll provide feedback faster I promise

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: @dagguh How do you keep up with maintaining such complex model inside your head? 
Though its good we have those tests <3 
@wyrzyk such comment/code might be  valuable inside test for others

PR #40, opened 2021-02-24
Introduce and use the experiment's API

wyrzyk:
Currently, we are dealing with the problem of rolling out multiple changes at once. It forced us to disable the whole feature. We should not make the same mistake in the future.

The first commit introduces the Experiment API. The second one shows how to use it on a simple bugfix.

### DIFF
--- a/docs/experiments.md
+++ b/docs/experiments.md
@@ -18,4 +18,4 @@ You can inject your implementation of `Experiment` via the `DualConnection` buil
 
 
 
-Experiments are short living. It's expected to remove it with the next releases.
+Experiments are short-lived. It's expected to remove it with the next releases.
--- a/docs/experiments.md
+++ b/docs/experiments.md
@@ -14,7 +14,7 @@ You can inject your implementation of `Experiment` via the `DualConnection` buil
 
 ## Current experiments
 
-none
+- `com.atlassian.db.replica.experiment.nativesql.on.replica` - controls "`Connection#nativeSQL` uses replica database" fix
 
 
 
--- a/docs/experiments.md
+++ b/docs/experiments.md
@@ -1,0 +1,21 @@
+## Experiments
+
+`Experiments` is an API to change the behaviour of the library [on the fly](https://martinfowler.com/articles/feature-toggles.html).
+The API is transparent, you need to `opt-in` to take control over the change.
+
+### Why
+
+This technique allows for safer rollouts of features and bugfixes than just bumping a library.
+
+### How
+
+You can inject your implementation of `Experiment` via the `DualConnection` builder.
+
+
+## Current experiments
+
+none
+
+
+
+Experiments are short living. It's expected to remove it with the next releases.
--- a/src/main/java/com/atlassian/db/replica/api/DualConnection.java
+++ b/src/main/java/com/atlassian/db/replica/api/DualConnection.java
@@ -3,7 +3,14 @@ package com.atlassian.db.replica.api;
 import com.atlassian.db.replica.api.circuitbreaker.BreakerState;
 import com.atlassian.db.replica.api.reason.Reason;
 import com.atlassian.db.replica.api.state.NoOpStateListener;
-import com.atlassian.db.replica.internal.*;
+import com.atlassian.db.replica.internal.Experiments;
+import com.atlassian.db.replica.internal.ForwardCall;
+import com.atlassian.db.replica.internal.ReadReplicaUnsupportedOperationException;
+import com.atlassian.db.replica.internal.ReplicaCallableStatement;
+import com.atlassian.db.replica.internal.ReplicaConnectionProvider;
+import com.atlassian.db.replica.internal.ReplicaPreparedStatement;
+import com.atlassian.db.replica.internal.ReplicaStatement;
+import com.atlassian.db.replica.internal.RouteDecisionBuilder;
 import com.atlassian.db.replica.internal.circuitbreaker.BreakOnNotSupportedOperations;
 import com.atlassian.db.replica.internal.circuitbreaker.BreakerConnection;
 import com.atlassian.db.replica.internal.circuitbreaker.BreakerHandler;
@@ -14,8 +21,29 @@ import com.atlassian.db.replica.spi.ReplicaConsistency;
 import com.atlassian.db.replica.spi.circuitbreaker.CircuitBreaker;
 import com.atlassian.db.replica.spi.state.StateListener;
 
-import java.sql.*;
-import java.util.*;
+import java.sql.Array;
+import java.sql.Blob;
+import java.sql.CallableStatement;
+import java.sql.ClientInfoStatus;
+import java.sql.Clob;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.NClob;
+import java.sql.PreparedStatement;
+import java.sql.SQLClientInfoException;
+import java.sql.SQLException;
+import java.sql.SQLWarning;
+import java.sql.SQLXML;
+import java.sql.Savepoint;
+import java.sql.Statement;
+import java.sql.Struct;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.Executor;
 
 /**
@@ -79,8 +107,13 @@ public final class DualConnection implements Connection {
     @Override
     public String nativeSQL(String sql) throws SQLException {
         checkClosed();
-        return connectionProvider.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL).sql(sql)).nativeSQL(
-            sql);
+        if (experiments.isEnabled("com.atlassian.db.replica.experiment.nativesql.on.replica")) {
+            return connectionProvider.getReadConnection(new RouteDecisionBuilder(Reason.RO_API_CALL).sql(sql)).nativeSQL(
+                sql);
+        } else {
+            return connectionProvider.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL).sql(sql)).nativeSQL(
+                sql);
+        }
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


docs/experiments.md
```diff
@@ -0,0 +1,21 @@
...
+You can inject your implementation of `Experiment` via the `DualConnection` builder.
+
+
+## Current experiments
+
+- `com.atlassian.db.replica.experiment.nativesql.on.replica` - controls "`Connection#nativeSQL` uses replica database" fix
+
+
+
+Experiments are short living. It's expected to remove it with the next releases.
```
dagguh, line 21:
```suggestion
Experiments are short-lived. It's expected to remove it with the next releases.
```

src/main/java/com/atlassian/db/replica/internal/Experiments.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class Experiments {
+    private final Map<String, Experiment> experimentsMap;
+
+    public Experiments(Collection<Experiment> experiments) {
+        final Map<String, Experiment> experimentsByKey = new HashMap<>();
+        for (Experiment experiment : experiments) {
+            experimentsByKey.put(experiment.getKey(), experiment);
```
dagguh, line 15: So we expect `Experiment.getKey` to be immutable

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -75,8 +107,13 @@ public CallableStatement prepareCall(String sql) throws SQLException {
     @Override
     public String nativeSQL(String sql) throws SQLException {
         checkClosed();
-        return connectionProvider.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL).sql(sql)).nativeSQL(
-            sql);
+        if (experiments.isEnabled("com.atlassian.db.replica.experiment.nativesql.on.replica")) {
```
dagguh, line 110: This is a new behavior the consumer wants to rely on. It's de facto API. So it's better make it explicit. And if a magic string is API, it's better to turn it to a method. Like `Experiments.shouldRunNativeSqlOnReplica()` with a default impl of `false`. Then after it's deemed stable, the SPI can remain as is without changing the default. The consumer can just hardcode it to `true` on their side. We can add Javadoc saying that `true` was well tested. Eventually, in a major release we can change the default to `true` or remove the toggle. But that's not necessary neither in the short term nor in the long term.

- [ ] reconsider magic string contract

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/api/DualConnection.java
wyrzyk: > This is a new behavior the consumer wants to rely on.

Experiments by definition are short-lived. We want to get rid of them as fast as we gain the confidence it works well. The only contract I would give for them is to be there for one release. It may and should be cleaned up with the next release. By changing the contracts from strings to methods, we can't remove the flags until the major release. The API will only grow.

Keeping the toggles that long would change the project into a maintenance nightmare with tens of flags that may manipulate the state.

I imagine this API like a zero-downtime upgrade feature. You can bump the library safely and control all the changes (revert if needed), but you can do it only from one version to the next version. I would not commit to providing an upgrade path for all features up to a major release.

dagguh: There's no "short-lived" API in SemVer. If the toggles would actually become a maintenance nightmare it just means we need a major release more often.

wyrzyk: `Experiment` is API. The behaviour for the changes behind the flags is not. Unfortunately, I have no better idea not to promise it will always work (especially there's no such use case. You either need it for rollouts or you don't need it at all. We don't need the flag after it's stable).

dagguh: Then any experiment can be broken at any lib bump. If that's ok by the consumers, then we can proceed

wyrzyk: > Then any experiment can be broken at any lib bump. If that's ok by the consumers, then we can proceed

Yes. Any lib bump can introduce errors. It's the standard way libs operate. `Experiment` is for safe bumps. Customers can also have other (better) ways for updates (OSGI for example), unfortunately, we don't use it.

dagguh: The consumer will make a typo in the magic string or will use a string that's no longer supported. It will not be detected at bump time or at compile time. It *might* get caught in runtime, but it can also remain hidden, because it won't throw exceptions.

Experiments will be used (for a short timer or not). If they don't affect public API (circumventing SemVer), then we'd need to maintain a separate list of supported experiments and when they were introduced and when they were dropped. `EXPERIMENT_LOG.md`?

I don't see why these new problems would be smaller than just releasing a major bump.

dagguh: Also see [Martin Fowler's `FeatureDecisions` example](https://www.martinfowler.com/articles/feature-toggles.html#De-couplingDecisionPointsFromDecisionLogic). It's much more powerful (and discoverable) than a `Map<String, boolean>`.

wyrzyk:
@dagguh The same problem, different approach (https://github.com/atlassian-labs/db-replica/pull/52). WDYT?

docs/experiments.md
```diff
@@ -0,0 +1,21 @@
+## Experiments
+
+`Experiments` is an API to change the behaviour of the library [on the fly](https://martinfowler.com/articles/feature-toggles.html).
```
afaruga-atlassian, line 3: Experiments look exactly as an approach to[hypothesis-driven-development](https://www.thoughtworks.com/insights/articles/how-implement-hypothesis-driven-development)
and FF are the RealTime solution to manipulate the hypothesis. 

Hypothesis still maybe controlled manually / automatically (if some exceptions-> rollback(aka. circuit breaker for new hypothesis)

src/test/java/com/atlassian/db/replica/internal/ExperimentsTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+import java.util.Collections;
+
+import static com.google.common.collect.ImmutableList.of;
+
+public class ExperimentsTest {
+
+    @Test
+    public void shouldBeEnabledByDefault() {
```
afaruga-atlassian, line 13: For me is's unnatural. Experiment should be disabled by default, as its experiment

src/test/java/com/atlassian/db/replica/internal/StaticExperiment.java
```diff
@@ -0,0 +1,24 @@
+package com.atlassian.db.replica.internal;
+
+import com.atlassian.db.replica.spi.Experiment;
+
+public final class StaticExperiment implements Experiment {
```
afaruga-atlassian, line 5: I guess `DynamicExperiment` or `FFDriverExperiment` would be implemented in the product?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
afaruga-atlassian: For consumer mistakes, and referring to the nature of [hypothesis-driven-development](https://www.thoughtworks.com/insights/articles/how-implement-hypothesis-driven-development) I'd consider adding automatic rollback of experiment/hypothesis, supported by circuit breaker - too many exceptions ? rolled it back automatically, or give a customer a way to skip automatic rollback, but not by default.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -75,8 +107,13 @@ public CallableStatement prepareCall(String sql) throws SQLException {
...
+            return connectionProvider.getReadConnection(new RouteDecisionBuilder(Reason.RO_API_CALL).sql(sql)).nativeSQL(
+                sql);
+        } else {
+            return connectionProvider.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL).sql(sql)).nativeSQL(
```
afaruga-atlassian, line 114: How do you see adding an experiment in ConnectionState? It looks like it will have to be injected along the way through all the classes in between, what require unproductive changes in unit tests.

StaticContext of Experiment would be a solution for it (it has it disadvantages)
I imaging that you would provide Experiment impl / configuration in DualConnectionBuilder as it is currently.

Then in any code in llb you could add:
`Experiement.getExperiment("this.is.just.example")`

Static code requires to have static mocks for configuration, what is bad, but doable.
On the other side, devloop is pretty fast and temporary changes in unit tests won't be a problem.

I don't have strong opinion, it looks like YAGNI.
Or maybe you had in mind that Experiments shouldn't be used in lol lvl internals?

src/test/java/com/atlassian/db/replica/internal/ExperimentsTest.java
wyrzyk: This is a library. We shouldn't force all users of the library to use experiments API. When you bump a library, your default expectation is to see the effect of changes or to have to find an experiment's key and switch it?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
wyrzyk: > It looks like it will have to be injected along the way through all the classes in between, what require unproductive changes in unit tests.

If injecting a new parameter will cause unit tests changes, then we have bad unit tests, and we should refactor them instead of avoiding changes.

> StaticContext of Experiment would be a solution for it

`static` is never a good choice. This approach would mask the problem with bad unit tests and will make them worse (harder to test, harder to run in parallel, possible flakiness).

afaruga-atlassian (approved):

wyrzyk:
Closing the PR. There's a better approach to solve the problem of unsafe-rollouts https://github.com/atlassian-labs/db-replica/pull/52.

Issue #41, opened 2021-02-26
org.postgresql.util.PSQLException: Cannot change transaction isolation level in the middle of a transaction.


wyrzyk:
```
Caused by: org.postgresql.util.PSQLException: Cannot change transaction isolation level in the middle of a transaction.
	at org.postgresql.jdbc.PgConnection.setTransactionIsolation(PgConnection.java:948)
	at com.atlassian.db.replica.internal.ConnectionParameters.initialize(ConnectionParameters.java:23)
	at com.atlassian.db.replica.internal.state.ConnectionState.prepareReadConnection(ConnectionState.java:187)
	at com.atlassian.db.replica.internal.state.ConnectionState.getReadConnection(ConnectionState.java:104)
	at com.atlassian.db.replica.internal.ReplicaConnectionProvider.getReadConnection(ReplicaConnectionProvider.java:43)
	at com.atlassian.db.replica.internal.ReplicaStatement$1.create(ReplicaStatement.java:32)
	at com.atlassian.db.replica.internal.ReplicaStatement$1.create(ReplicaStatement.java:29)
	at com.atlassian.db.replica.internal.LazyReference.get(LazyReference.java:20)
	... 54 more

```

wyrzyk:
Fixed in `0.1.27`

PR #42, opened 2021-02-26
Initialize connection once

wyrzyk:
Fix for #41

### DIFF
--- a/src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
@@ -5,6 +5,8 @@ import java.sql.SQLException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
 
 public final class ConnectionParameters {
 
@@ -36,7 +38,11 @@ public final class ConnectionParameters {
         }
     }
 
-    public void setTransactionIsolation(Integer transactionIsolation) {
+    public void setTransactionIsolation(
+        Supplier<Optional<Connection>> currentConnection,
+        Integer transactionIsolation
+    ) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setTransactionIsolation(transactionIsolation));
         this.transactionIsolation = transactionIsolation;
     }
 
@@ -44,7 +50,11 @@ public final class ConnectionParameters {
         return this.transactionIsolation;
     }
 
-    public void setAutoCommit(Boolean autoCommit) {
+    public void setAutoCommit(
+        Supplier<Optional<Connection>> currentConnection,
+        Boolean autoCommit
+    ) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setAutoCommit(autoCommit));
         this.isAutoCommit = autoCommit;
     }
 
@@ -56,7 +66,8 @@ public final class ConnectionParameters {
         return catalog;
     }
 
-    public void setCatalog(String catalog) {
+    public void setCatalog(Supplier<Optional<Connection>> currentConnection, String catalog) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setCatalog(catalog));
         this.catalog = catalog;
     }
 
@@ -64,7 +75,11 @@ public final class ConnectionParameters {
         return typeMap == null ? Collections.emptyMap() : new HashMap<>(typeMap);
     }
 
-    public void setTypeMap(Map<String, Class<?>> typeMap) {
+    public void setTypeMap(
+        Supplier<Optional<Connection>> currentConnection,
+        Map<String, Class<?>> typeMap
+    ) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setTypeMap(typeMap));
         this.typeMap = typeMap;
     }
 
@@ -72,7 +87,11 @@ public final class ConnectionParameters {
         return holdability;
     }
 
-    public void setHoldability(Integer holdability) {
+    public void setHoldability(
+        Supplier<Optional<Connection>> currentConnection,
+        Integer holdability
+    ) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setHoldability(holdability));
         this.holdability = holdability;
     }
 
@@ -80,8 +99,19 @@ public final class ConnectionParameters {
         return readOnly != null && readOnly;
     }
 
-    public void setReadOnly(boolean readOnly) {
+    public void setReadOnly(Supplier<Optional<Connection>> currentConnection, boolean readOnly) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setReadOnly(readOnly));
         this.readOnly = readOnly;
+    }
+
+    private void executeIfPresent(
+        Supplier<Optional<Connection>> currentConnection,
+        ConnectionOperation operation
+    ) throws SQLException {
+        final Optional<Connection> connection = currentConnection.get();
+        if (connection.isPresent()) {
+            operation.accept(connection.get());
+        }
     }
 
     @Override
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
@@ -43,8 +43,8 @@ public class ReplicaConnectionProvider implements AutoCloseable {
         return state.getReadConnection(decisionBuilder);
     }
 
-    public void setTransactionIsolation(Integer transactionIsolation) {
-        parameters.setTransactionIsolation(transactionIsolation);
+    public void setTransactionIsolation(Integer transactionIsolation) throws SQLException {
+        parameters.setTransactionIsolation(state::getConnection, transactionIsolation);
     }
 
     public int getTransactionIsolation() throws SQLException {
@@ -57,11 +57,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
 
     public void setAutoCommit(Boolean autoCommit) throws SQLException {
         final boolean autoCommitBefore = getAutoCommit();
-        parameters.setAutoCommit(autoCommit);
-        final Optional<Connection> connection = state.getConnection();
-        if (connection.isPresent()) {
-            connection.get().setAutoCommit(autoCommit);
-        }
+        parameters.setAutoCommit(state::getConnection, autoCommit);
         if (autoCommitBefore != getAutoCommit()) {
             recordCommit(autoCommitBefore);
         }
@@ -80,7 +76,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     }
 
     public void setReadOnly(boolean readOnly) throws SQLException {
-        parameters.setReadOnly(readOnly);
+        parameters.setReadOnly(state::getConnection, readOnly);
     }
 
     public String getCatalog() {
@@ -88,11 +84,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     }
 
     public void setCatalog(String catalog) throws SQLException {
-        parameters.setCatalog(catalog);
-        final Optional<Connection> connection = state.getConnection();
-        if (connection.isPresent()) {
-            connection.get().setCatalog(catalog);
-        }
+        parameters.setCatalog(state::getConnection, catalog);
     }
 
     public Map<String, Class<?>> getTypeMap() {
@@ -100,11 +92,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     }
 
     public void setTypeMap(Map<String, Class<?>> typeMap) throws SQLException {
-        parameters.setTypeMap(typeMap);
-        final Optional<Connection> connection = state.getConnection();
-        if (connection.isPresent()) {
-            connection.get().setTypeMap(typeMap);
-        }
+        parameters.setTypeMap(state::getConnection, typeMap);
     }
 
     public Integer getHoldability() throws SQLException {
@@ -112,11 +100,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     }
 
     public void setHoldability(Integer holdability) throws SQLException {
-        parameters.setHoldability(holdability);
-        final Optional<Connection> connection = state.getConnection();
-        if (connection.isPresent()) {
-            connection.get().setHoldability(holdability);
-        }
+        parameters.setHoldability(state::getConnection, holdability);
     }
 
     public SQLWarning getWarning() throws SQLException {
@@ -151,10 +135,6 @@ public class ReplicaConnectionProvider implements AutoCloseable {
         } else {
             return currentConnection.isWrapperFor(iface);
         }
-    }
-
-    public boolean hasWriteConnection() {
-        return this.state.hasWriteConnection();
     }
 
     public State getState() {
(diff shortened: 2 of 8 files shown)


src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -36,6 +48,12 @@ public void initialize(Connection connection) throws SQLException {
         }
     }
 
+    private boolean isInitialized(Connection connection) {
+        return initializedConnections.stream().map(WeakReference::get).filter(Objects::nonNull).collect(
```
wyrzyk, line 52: It's too slow. Will not pass the performance test.

dwatl (approved):

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
```diff
@@ -79,6 +79,22 @@ public void shouldRunNextValOnMainDatabase() throws SQLException {
         }
     }
 
+    @Test
+    public void shouldChangeTransactionIsolationLevel() throws SQLException {
```
dwatl, line 83: Why should it change level, when?
Could we put an actual contract in the name, I can read from test impl that it's changing level - but why?

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
```diff
@@ -79,6 +79,22 @@ public void shouldRunNextValOnMainDatabase() throws SQLException {
...
+        try (PostgresConnectionProvider connectionProvider = new PostgresConnectionProvider()) {
+            final Connection connection = DualConnection.builder(
+                connectionProvider,
+                CircularConsistency.permanentConsistency().build()
+            ).build();
+
+            connection.setTransactionIsolation(Connection.TRANSACTION_READ_UNCOMMITTED);
+            connection.setAutoCommit(false);
```
dwatl, line 91: What if autoCommit=true? Will it still work fine?

dwatl (approved):

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
wyrzyk: Yes. It just reproduces exactly the issue we seen on the production. There's a more generic unit test that targets the root cause.

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
wyrzyk: Changing level is part of the connection's API. It should change the level when you call the API. Maybe more precise would be shluldNotFailWhenChangingTransactionIsolationLevel WDYT?

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -14,8 +12,13 @@
     private String catalog;
     private Map<String, Class<?>> typeMap;
     private Integer holdability;
+    private final Set<Connection> initializedConnections = Collections.newSetFromMap(new WeakHashMap<>());
```
dagguh, line 15: There's `State.NOT_INITIALISED`. Is that the same thing? Also `ConnectionState` has calls like `readConnection.isInitialized()`. Is that related?

It starts to be hard to think about. `DualConnection` has some immutable **state**. `ReplicaConnectionProvider` has the  `ConnectionParameters` *and* `ConnectionState` as its state. Then `ConnectionState` has the `ConnectionParameters` *again* and it returns a different `State`.

It seems we lost the OOP way and went back to data structures and procedures. This grows the code, because reuse is limited, because different procedures work on different data structures rather than on the same objects.

dagguh: I think we can still go a little further in this style to get replicas running on prod. But soon after, the tech debt will need to be paid off.

wyrzyk: > There's State.NOT_INITIALISED. Is that the same thing?

No. It's the state of the dual connection. It means no connection was initialized. We use real connections internally. We may initialise them multiple times (first replica, then main connection, because replica was inconsistent, and again replica as it's already consistent ....). Both are states but are not related.

`readConnection.isInitialized()` it's from a lazy connection. Now it means only it's obtained from the pool. Maybe it's a good idea to initialize the parameters here and return only fully-initialized connections from the pool.

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -14,8 +12,13 @@
...
 
     public void initialize(Connection connection) throws SQLException {
+        if (initializedConnections.contains(connection)) {
```
afaruga-atlassian, line 18: What about changing isolation level for an already initialized connection ? Not sure if that is possible but reading through unit tests this come to my mind.

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
wyrzyk: @dagguh, I refactored it. Is it better now?

wyrzyk: > It seems we lost the OOP way and went back to data structures and procedures.

Can you elaborate?

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
```diff
@@ -79,6 +79,22 @@ public void shouldRunNextValOnMainDatabase() throws SQLException {
...
+            connection.prepareStatement("SELECT 1;").executeQuery();
+            connection.prepareStatement("SELECT 1;").executeQuery();
+            connection.commit();
+        }
```
afaruga-atlassian, line 95: Does this test how the connection tried to be initialized twice with changed parameters?

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
dagguh: > @dagguh, I refactored it. Is it better now?

Much simpler, whoa! My 🐒🧠 can track it now 😉

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
wyrzyk: I think it's the reason for another problem we see on prod. I'm working on it 😄 .

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
dagguh: In OOP, objects protect their state via encapsulation and protect the control over them by exposing only behaviors (not commands like setters). For example, if `ConnectionState` is a state of a connection, it should be encapsulated in the connection. It also steals behaviors from `State` and `Connection` by issuing commands to them and by digging in their internals.

wyrzyk: `State` has no behaviours. It's an enum. `ConnectionState` exposes the current `state` of DualConnection. To obtain the current state, it needs to verify what raw `connections` are initialized (as the `state` describes it). It's also the only point to change the state as it encapsulates the raw connections. It would lose control over changing and retrieving the state without it. Currently, ConnectionState can:
- control the state --> "issuing commands on raw connections."
- retrieve the state--> "steals behaviors from State"

`ConnectionState` is encapsulated in `DualConnection` (unfortunately, there's also wrongly named mediator `ReplicaConnectionProvider`, but it's something to refactor in the future). 

I still don't get what's not OO here. Maybe you mix DualConnection (smart proxy over real connections) with raw connections (state of the DualConnection).

wyrzyk: We could extract control over the state from `ConnectionState` to a different class. Would it make it more OO?

(there are also some methods that seem to be unrelated to the ConnectinState that we can remove `hasWriteConnection`, `getDecision`)

wyrzyk: It would reduce ConnectionState to 

```
 public State getState() {
        if (isClosed != null && isClosed) {
            return CLOSED;
        } else {
            final boolean readReady = readConnection.isInitialized();
            final boolean writeReady = writeConnection.isInitialized();
            if (!readReady && !writeReady) {
                return NOT_INITIALISED;
            } else if (!replicaConsistent && writeReady) {
                return COMMITED_MAIN;
            } else if (writeReady) {
                return MAIN;
            } else {
                return REPLICA;
            }
        }
    }
```

dagguh: These "state" classes look like a bigger object was cut into pieces, but not with a clean edge. It's like they still have leftover connections between them and they expose fields and methods to each other to keep working. By doing that, they leak it to others too. Objects have behaviors and protect their internals, so enums and DTOs are not objects (all state no behavior).

dagguh: > It would reduce ConnectionState to

Yeah, reducing or entirely dissolving a "state" object would definitely help encapsulation

wyrzyk: > These "state" classes look like a bigger object was cut into pieces, but not with a clean edge.

Yes, ConnectionState cut it from multiple places. It's not the final shape it should be. 
So I think we have a consensus to fix it by extracting part of the ConnectionState to more appropriate places.
But I don't get this part:

> dissolving a "state" object

Can you elaborate? We didn't have an explicit `state` before, so all the checks were scattered here and there. State reduces it to one place. You no longer decide how to act depending on random ifs that checks what connection is initialized, it's all unified in one place.

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
wyrzyk: Ok. This one was covered by setting the current connection state in another place, but `setReadOnly` was not and it was causing the problems. I refactored the code to set the properties in one place and to cover setReadOnly and other parameters.

wyrzyk: I put it in the same PR as it seems to be related.

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
wyrzyk: I plan to use some of my 20% for refactoring. As you mentioned, let's make it green and then we can continue. For me, the priorities are:
- make it work 😉 ,
- introduce a stable API (1.0.0 release), 
- refactor internals.

Let us start with the refactor that may affect API.

@dagguh Thanks for all the feedback and for looking at the code with fresh eyes. It went a long way from the in-product spike to the lib. It went from ultra-complex code to less complex, but still, it requires improvement. I'd appreciate it if you start to refactor to show proposed changes. It would be much easier to talk about the changes.

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -36,15 +38,23 @@ public void initialize(Connection connection) throws SQLException {
         }
     }
 
-    public void setTransactionIsolation(Integer transactionIsolation) {
+    public void setTransactionIsolation(
+        Supplier<Optional<Connection>> currentConnection,
```
wyrzyk, line 42: In the current form, `ConnectionParameters` could contain the `state` and set the transaction if any raw connection initialized. I could refactor it now, but `state` needs to be refactored anyway. I don't want to enter the rabbit-hole of the refactoring 😨 .

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
wyrzyk: Yes.

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
dagguh: Gathering control from all around the codebase into a few objects was indeed a good and needed move. It remained procedural, because "explicit state" is a data structure. And with objects we want to hide them, exposing only behaviors. I'll open a refactor PR next week 👍

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -36,15 +38,23 @@ public void initialize(Connection connection) throws SQLException {
...
+        Integer transactionIsolation
+    ) throws SQLException {
+        executeIfPresent(currentConnection, connection -> connection.setTransactionIsolation(transactionIsolation));
         this.transactionIsolation = transactionIsolation;
```
afaruga-atlassian, line 46: Can transactionIsolation in connection and connection parameters can go away and differ each other?

wyrzyk: I don't understand the questions, but I'll describe what happens in the code here. We set the parameters. Parameters are for database connections that are not initialised yet. We need to store them for future use. Sometimes we already have a connection. Then we need to set the parameter directly. We still have to store the parameter for future use (we change connections dynamically).

Here if there's no current connection. We store a parameter.
If there's a connection and it will fail to `setTransactionIsolation`, it'll throw an exception and not store it for the future.

Does it answer the questions?

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
dagguh: Fair call

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
dagguh: It's a similar situation as with `StatementOperation`, which is used to defer calls to the _current_ `Statement`

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -153,10 +137,6 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
         }
     }
 
-    public boolean hasWriteConnection() {
```
dagguh, line 156: Bye bye 👋

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -49,7 +51,9 @@ public Connection create() throws SQLException {
     private final DecisionAwareReference<Connection> writeConnection = new DecisionAwareReference<Connection>() {
         @Override
         public Connection create() throws SQLException {
-            return connectionProvider.getMainConnection();
+            final Connection mainConnection = connectionProvider.getMainConnection();
+            parameters.initialize(mainConnection);
```
dagguh, line 55: Initializing parameters when initializing connection sounds like a good idea 👍

src/test/java/com/atlassian/db/replica/it/DualConnectionIT.java
dagguh: It doesn't assert on the new isolation level. So indeed `shouldNotFailWhen...` or the assertion.

dagguh (approved):

dagguh (approved):

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
afaruga-atlassian: @wyrzyk thanks for explnation it covered my question.

> We set the parameters. Parameters are for database connections that are not initialised ye
Can we isolate initialized from none-initialized connections?

PR #47, opened 2021-03-01
Harden Connection#close

wyrzyk:
We ensure `ConnectionParameters` is in sync with the currently used connection, so it should not matter if we check the state from
the connection or parameters, but checking it directly on
a connection seems to be less error-prone.

### DIFF
--- a/src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
+++ b/src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
@@ -210,7 +210,7 @@ public final class ConnectionState {
             } catch (Exception e) {
                 warnings.saveWarning(new SQLWarning(e));
             }
-            if (parameters.isReadOnly()) {
+            if (connection.isReadOnly()) {
                 connection.setAutoCommit(true);
                 connection.setReadOnly(false);
             }
--- a/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
+++ b/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
@@ -1420,18 +1420,17 @@ public class TestDualConnection {
 
     @Test
     public void shouldResetReadOnlyModeBeforeReleasingConnection() throws SQLException {
-        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
-        final Connection connection = DualConnection.builder(
-            connectionProvider,
+        final Connection connection = new ConnectionMock();
+        final SingleConnectionProvider singleConnectionProvider = new SingleConnectionProvider(connection);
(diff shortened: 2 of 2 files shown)


src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -210,7 +210,7 @@ private void closeConnection(
             } catch (Exception e) {
                 warnings.saveWarning(new SQLWarning(e));
             }
-            if (parameters.isReadOnly()) {
+            if (connection.isReadOnly()) {
```
dagguh, line 213: The `connection` <-> `parameters` dualism require sync. It'd be better if only one of them remained in existence.
I'll take a stab at it in the OOP refactor.

dagguh: Red `TestDualConnection.shouldResetReadOnlyModeBeforeReleasingConnection`:
```
Wanted but not invoked:
autoCommitAwareConnection.setReadOnly(false);
-> at com.atlassian.db.replica.api.TestDualConnection.shouldResetReadOnlyModeBeforeReleasingConnection(TestDualConnection.java:1435)

However, there were exactly 3 interactions with this mock:
autoCommitAwareConnection.getWarnings();
-> at com.atlassian.db.replica.internal.state.ConnectionState.closeConnection(ConnectionState.java:209)

autoCommitAwareConnection.isReadOnly();
-> at com.atlassian.db.replica.internal.state.ConnectionState.closeConnection(ConnectionState.java:213)

autoCommitAwareConnection.close();
-> at com.atlassian.db.replica.internal.state.ConnectionState.closeConnection(ConnectionState.java:217)
```

Hardening affected the mocking?

dagguh (approved):

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: Yes. The mock didn't have to have this state in sync :). So the test needs to improve with the code :).

wyrzyk: >  It'd be better if only one of them remained in existence.

🤞 
It will be awesome if it's possible. `parameters` keep more data than connection. It also knows if the property has been changed.

dagguh (dismissed):

dagguh (approved):
That [force-push](https://github.com/atlassian-labs/db-replica/compare/a2d5c3fcb06088c43d55b9532454557adf71b6cd..76bf2481035a22275cc8b5bc271d73f498f8b10f) was just a rebase onto upstream.

PR #48, opened 2021-03-01
JCES-1896 Adding isWrite information to API

afaruga-atlassian:


src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,14 +42,31 @@ public Reason getReason() {
         return Optional.ofNullable(cause);
     }
 
+    /**
+     * @return information whether reason was caused by write sql operation.
+     */
+    public boolean isWrite(){
+        return WRITE_OPERATION.equals(reason) ||
```
afaruga-atlassian, line 49: I want to start discussion here.
I think this answer will work in majority of the cases but not all.

I think there is no way to have a class like SQLQuery.isSQLModifyDatabase(sql) - though I have to say It was tempting for me in place to think about it, it would be extracted from `ReplicaStatement`
with methods like
`sqlFunction.isFunctionCall(sql) || isUpdate(sql) || isDelete(sql);`
But why i didn't do that?
Cause this is not enough. I think SQL + statementType + still there are chances that 
`ReplicaPreparedStatement.execute() ` will run either READ or Write operation.

I wonder whether we should introduce localState(localField) in `ReplicaStatement` that would determine whether query isReadOnly or not.
(it might be easier to determine whether query is ReadOnly, or whether isWrite).
@wyrzyk How would you approach it ?

afaruga-atlassian: P.S. 
Initially I thought about extending `Reason` and adding one more boolean field isWrite.

But some of states might be either write or read or changed. I think initliallty `Reason` was created present other dimensions. On from those dimensions you cannot ultimately judge whether Reason was write or read. 

Am I right ? 
I'm asking for suggestions.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,14 +42,31 @@ public Reason getReason() {
         return Optional.ofNullable(cause);
     }
 
+    /**
+     * @return information whether reason was caused by write sql operation.
```
wyrzyk, line 46: A JavaDoc ❤️

wyrzyk, line 46:
```suggestion
     * @return Information whether a reason was caused by write SQL operation.
```

src/test/java/com/atlassian/db/replica/api/reason/TestRouteDecision.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+public class TestRouteDecision {
```
dwatl, line 14: Actually, what do we test here and why?

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -57,9 +78,9 @@ public int hashCode() {
     @Override
     public String toString() {
         return "RouteDecision{" +
-            "reason=" + reason +
-            ", sql='" + sql + '\'' +
-            ", cause=" + cause +
-            '}';
+                "reason=" + reason +
```
dwatl, line 81: :/

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,14 +42,31 @@ public Reason getReason() {
...
+     * @return information whether reason was caused by write sql operation.
+     */
+    public boolean isWrite(){
+        return WRITE_OPERATION.equals(reason) ||
+                LOCK.equals(reason) ||
+                RW_API_CALL.equals(reason);
+    }
+
+    /**
+     * @return whether sql was run on Main
```
dwatl, line 55: Was implies it was post-fact, is it true?

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
wyrzyk: We always know if the query is a read or write operation. That's the library's main goal: to split read and write operations between the main database and replicas.

Extending `Reason` with `isWrite` seems to be a good call.

> But some of states might be either write or read or changed

What do you mean? Can you elaborate?

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,14 +42,31 @@ public Reason getReason() {
...
+     */
+    public boolean isRunOnMain() {
```
wyrzyk, line 57: Any reason to add this?

src/test/java/com/atlassian/db/replica/api/reason/TestRouteDecision.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private static final RouteDecision READ_DECISION = new RouteDecision("select * from somewhere", Reason.READ_OPERATION, null);
```
wyrzyk, line 16: Nit. You could reuse `com.atlassian.db.replica.api.Queries.SIMPLE_QUERY`.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: @wyrzyk I've created a draft PR especially to elaborate my thinking here: 
Please take a look. I've added a bunch of todo/comments overthere, looking for a best place to decide whether query is read/write, cause for me there are couple of places when actually we don;t know, or for safety reasons we treat querys as writes, while based on SQL they aren't writes.

https://github.com/atlassian-labs/db-replica/pull/49

wyrzyk: > or for safety reasons we treat querys as writes, while based on SQL they aren't writes.

Yes. That's true. Some of the places are tracked in GH issues. I guess we're going to resolve some of these issues when we gain better observability.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
dagguh: It's semantic. The main point of the class is to *decide* to go to main or not. From API it seems that `RouteDecision` has a "Reason", but doesn't even know what the decision is.
If `isWrite` will go here, then it would also add symmetry.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
dagguh: I would define a "write" like this:
```suggestion
     * @return true if the database call would always fail on a read replica
```

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: I don't have strong opinion here, it just looks a bit nicer to have all of this inside RouteDecision.
I've failed adding `write` field into `Reason`. So this looks to be nicer.

wyrzyk: So we are moving it from 'Reason`? I would like to make sure we don't create two alternative ways to reach the same data. The reason to mix Reason and `isWrite` in one object was to not allow broken states like `RW_OPeration' isWrite=false on compile time.

afaruga-atlassian: Good catch!

- [ if isRunOnMain() moved, delete it from `Reason`]

dagguh: If `Reason.isRunOnMain` becomes package-private, then we get both the internal consistency and one way to call the API.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: > true if the database call would always fail on a read replica
@dagguh interesting suggestion to define what is Write 

Are we sure about always? or always on specific db-replica version?

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: I'll prefer to migrate internals to also use  `RouteDecision.isRunOnMain()` and have 1 source of truth

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: I'm afraid we don't have implemented logic that will determine this.

dagguh: Always = not-transient, retrying will not help.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: At the end in draft it turns out that there are only two types of Writes operations (`RW_API_CALL` , `WRITE_OPERATION`
Rest is read, but might be treated as Write due to circumstances with connection state, transaction lvl,etc.
@dagguh 
@wyrzyk

afaruga-atlassian:
I'm closing this PR on behalf of https://github.com/atlassian-labs/db-replica/pull/49

Issue #37, opened 2021-02-24
`Connection#nativeSQL` doesn't need the main database connection


wyrzyk:
The method 'Converts the given SQL statement into the system's native SQL grammar.'. It seems the method can utilise a replica connection (it can use even inconsistent replica 🤔 ).

wyrzyk:
Fixed in `0.1.30`.

Issue #9, opened 2021-01-26
Maintain zero dependencies


dagguh:
Currently, the [`compileClasspath`](https://github.com/atlassian-labs/db-replica/blob/19b83af9838413c414cb64c238b7ff819082394b/gradle/dependency-locks/compileClasspath.lockfile) and the [`runtimeClasspath`](https://github.com/atlassian-labs/db-replica/blob/19b83af9838413c414cb64c238b7ff819082394b/gradle/dependency-locks/runtimeClasspath.lockfile) are empty.
It's a precious feature, because it means that `db-replica` does not generate transitive dependency conflicts. It also has lower network/runtime footprint for the consumers.

This feature can be maintained by checking it via a Gradle task.

PR #49, opened 2021-03-02
Issue/jces 1896 is write api sql draft

afaruga-atlassian:
JCES-1896 draft - I'm trying to find the ultimate answer for how to judge whether sql is read/write and where is best place (is there possibility to increase coverage by more readQueries moved to Replica). How SQL operation is treated in db-replica is based on statementType + sql + consistency. Is there a way to arbitrarily judge whether SQL is read/write?

### DIFF
--- a/src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
+++ b/src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
@@ -14,11 +14,21 @@ public final class RouteDecision {
     private final Reason reason;
     private final String sql;
     private final RouteDecision cause;
+    private final Boolean isWrite;
 
     public RouteDecision(String sql, Reason reason, RouteDecision cause) {
         this.sql = sql;
         this.reason = reason;
         this.cause = cause;
+        this.isWrite = null;
+    }
+
+    public RouteDecision(String sql, Reason reason, RouteDecision cause, Boolean isWrite) {
+        this.sql = sql;
+        this.reason = reason;
+        this.cause = cause;
+        this.isWrite = isWrite;
+
     }
 
     /**
@@ -42,13 +52,8 @@ public final class RouteDecision {
         return Optional.ofNullable(cause);
     }
 
-    /**
-     * @return information whether reason was caused by write sql operation.
-     */
-    public boolean isWrite(){
-        return WRITE_OPERATION.equals(reason) ||
-                LOCK.equals(reason) ||
-                RW_API_CALL.equals(reason);
+    public Boolean getWrite() {
+        return isWrite;
     }
 
     /**
@@ -58,21 +63,22 @@ public final class RouteDecision {
         return reason.isRunOnMain();
     }
 
-
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
         RouteDecision that = (RouteDecision) o;
-        return Objects.equals(reason, that.reason) && Objects.equals(
-                sql,
-                that.sql
-        ) && Objects.equals(cause, that.cause);
+        return Objects.equals(reason, that.reason) && Objects.equals(sql, that.sql) && Objects
+                .equals(cause, that.cause) && Objects.equals(isWrite, that.isWrite);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(reason, sql, cause);
+        return Objects.hash(reason, sql, cause, isWrite);
     }
 
     @Override
@@ -81,6 +87,7 @@ public final class RouteDecision {
                 "reason=" + reason +
                 ", sql='" + sql + '\'' +
                 ", cause=" + cause +
+                ", isWrite=" + isWrite +
                 '}';
     }
 }
--- a/src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
+++ b/src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
@@ -3,6 +3,10 @@ package com.atlassian.db.replica.api.reason;
 import java.util.Objects;
 import java.util.Optional;
 
+import static com.atlassian.db.replica.api.reason.Reason.LOCK;
+import static com.atlassian.db.replica.api.reason.Reason.RW_API_CALL;
+import static com.atlassian.db.replica.api.reason.Reason.WRITE_OPERATION;
+
 /**
  * Reveals details related to why, and which database will be used.
  */
@@ -38,14 +42,31 @@ public final class RouteDecision {
         return Optional.ofNullable(cause);
     }
 
+    /**
+     * @return information whether reason was caused by write sql operation.
+     */
+    public boolean isWrite(){
+        return WRITE_OPERATION.equals(reason) ||
+                LOCK.equals(reason) ||
+                RW_API_CALL.equals(reason);
+    }
+
+    /**
+     * @return whether sql was run on Main
+     */
+    public boolean isRunOnMain() {
+        return reason.isRunOnMain();
+    }
+
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         RouteDecision that = (RouteDecision) o;
         return Objects.equals(reason, that.reason) && Objects.equals(
-            sql,
-            that.sql
+                sql,
+                that.sql
         ) && Objects.equals(cause, that.cause);
     }
 
@@ -57,9 +78,9 @@ public final class RouteDecision {
     @Override
     public String toString() {
         return "RouteDecision{" +
-            "reason=" + reason +
-            ", sql='" + sql + '\'' +
-            ", cause=" + cause +
-            '}';
+                "reason=" + reason +
+                ", sql='" + sql + '\'' +
+                ", cause=" + cause +
+                '}';
     }
 }
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
@@ -25,7 +25,8 @@ public class ReplicaStatement implements Statement {
     private final ReplicaConsistency consistency;
     private final DatabaseCall databaseCall;
     private boolean isWriteOperation = true;
-    private final SqlFunction sqlFunction;
+
+    private SQLOpeartionType sqlOpeartionType;
     private final DecisionAwareReference<Statement> readStatement = new DecisionAwareReference<Statement>() {
         @Override
         public Statement create() throws Exception {
@@ -54,13 +55,13 @@ public class ReplicaStatement implements Statement {
         this.resultSetType = resultSetType;
         this.resultSetConcurrency = resultSetConcurrency;
         this.resultSetHoldability = resultSetHoldability;
-        this.sqlFunction = new SqlFunction(readOnlyFunctions);
+        this.sqlOpeartionType =  new SQLOpeartionType(new SqlFunction(readOnlyFunctions));;
     }
 
     @Override
     public ResultSet executeQuery(String sql) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(READ_OPERATION).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(READ_OPERATION).sql(sql).isWrite(false);
         final Statement statement = getReadStatement(decisionBuilder);
         return execute(() -> statement.executeQuery(sql), decisionBuilder.build());
     }
@@ -68,7 +69,7 @@ public class ReplicaStatement implements Statement {
     @Override
     public int executeUpdate(String sql) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql).isWrite(true);
         final Statement statement = getWriteStatement(decisionBuilder);
         return execute(
             () -> statement.executeUpdate(sql),
@@ -94,7 +95,8 @@ public class ReplicaStatement implements Statement {
     @Override
     public int getMaxFieldSize() throws SQLException {
         checkClosed();
-        return getWriteStatement(new RouteDecisionBuilder(RW_API_CALL)).getMaxFieldSize();
+        //todo this looks to be READ operation but run on write statement?
+        return getWriteStatement(new RouteDecisionBuilder(RW_API_CALL).isWrite(false)).getMaxFieldSize();
     }
 
     @Override
@@ -108,7 +110,8 @@ public class ReplicaStatement implements Statement {
     @Override
     public int getMaxRows() throws SQLException {
         checkClosed();
-        return getWriteStatement(new RouteDecisionBuilder(RW_API_CALL)).getMaxRows();
+        //todo this looks to be READ operation but run on write statement?
+        return getWriteStatement(new RouteDecisionBuilder(RW_API_CALL).isWrite(false)).getMaxRows();
     }
 
     @Override
@@ -176,11 +179,12 @@ public class ReplicaStatement implements Statement {
         checkClosed();
         final RouteDecisionBuilder decisionBuilder;
         final Statement statement;
-        if (sql.startsWith("set")) {
-            decisionBuilder = new RouteDecisionBuilder(READ_OPERATION).sql(sql);
+        if (sqlOpeartionType.isSQLSet(sql)) {
+            decisionBuilder = new RouteDecisionBuilder(READ_OPERATION).sql(sql).isWrite(false);
             statement = getReadStatement(decisionBuilder);
         } else {
             decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+            //todo not sure whether this is write sql, it's based on hidden contract that  execute(String sql) is run only by Write SQL
             statement = getWriteStatement(decisionBuilder);
         }
         return execute(
@@ -273,6 +277,7 @@ public class ReplicaStatement implements Statement {
     public int[] executeBatch() throws SQLException {
         checkClosed();
         final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL);
+        //todo is it write or read?
         final Statement statement = getWriteStatement(decisionBuilder);
         return execute(statement::executeBatch, decisionBuilder.build());
     }
@@ -280,6 +285,7 @@ public class ReplicaStatement implements Statement {
     @Override
     public Connection getConnection() throws SQLException {
         checkClosed();
+        //todo is it write or read?
         return connectionProvider.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL));
     }
 
@@ -309,7 +315,7 @@ public class ReplicaStatement implements Statement {
     @Override
     public int executeUpdate(String sql, int[] columnIndexes) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql).isWrite(true);
         final Statement statement = getWriteStatement(decisionBuilder);
         return execute(
             () -> statement.executeUpdate(sql, columnIndexes),
@@ -320,7 +326,7 @@ public class ReplicaStatement implements Statement {
     @Override
     public int executeUpdate(String sql, String[] columnNames) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql).isWrite(true);
         final Statement statement = getWriteStatement(decisionBuilder);
         return execute(
             () -> statement.executeUpdate(sql, columnNames),
@@ -331,7 +337,8 @@ public class ReplicaStatement implements Statement {
     @Override
     public boolean execute(String sql, int autoGeneratedKeys) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql).isWrite(true);
+        //todo is it write or read?
         final Statement statement = getWriteStatement(decisionBuilder);
         return execute(
             () -> statement.execute(sql, autoGeneratedKeys),
@@ -342,7 +349,7 @@ public class ReplicaStatement implements Statement {
     @Override
     public boolean execute(String sql, int[] columnIndexes) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql).isWrite(true);
         final Statement state = getWriteStatement(decisionBuilder);
         return execute(
             () -> state.execute(sql, columnIndexes),
@@ -353,11 +360,11 @@ public class ReplicaStatement implements Statement {
     @Override
     public boolean execute(String sql, String[] columnNames) throws SQLException {
         checkClosed();
-        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql);
+        final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL).sql(sql).isWrite(true);
         final Statement statement = getWriteStatement(decisionBuilder);
         return execute(
             () -> statement.execute(sql, columnNames),
-            decisionBuilder.build()
+            decisionBuilder.build() //todo this looks like best place to generically state whether query is Write/Read opeariton
         );
     }
 
@@ -554,41 +561,35 @@ public class ReplicaStatement implements Statement {
 
     public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) {
         if (connectionProvider.getState().equals(MAIN)) {
-            decisionBuilder.reason(MAIN_CONNECTION_REUSE);
+            decisionBuilder.reason(MAIN_CONNECTION_REUSE);//.isWrite(true) ? ;
+            // sth was readonly, but need to happen on Main because of inconsistency
+            // I think we don't have info whether this is read  / write
             connectionProvider.getStateDecision().ifPresent(decisionBuilder::cause);
+
+
             return prepareWriteStatement(decisionBuilder);
         }
         final String sql = decisionBuilder.getSql();
-        isWriteOperation = sqlFunction.isFunctionCall(sql) || isUpdate(sql) || isDelete(sql);
+        isWriteOperation = sqlOpeartionType.isWriteOpearation(sql);
         if (isWriteOperation) {
-            decisionBuilder.reason(WRITE_OPERATION);
+            decisionBuilder.reason(WRITE_OPERATION).isWrite(true);
             return prepareWriteStatement(decisionBuilder);
         }
 
-        if (isSelectForUpdate(sql)) {
-            decisionBuilder.reason(LOCK);
+        if (sqlOpeartionType.isSelectForUpdate(sql)) {
+            decisionBuilder.reason(LOCK).isWrite(true);
             return prepareWriteStatement(decisionBuilder);
         }
 
+        // todo what should happen here UNKNOWN SQL operation type ?
         setCurrentStatement(getCurrentStatement() != null ? getCurrentStatement() : readStatement.get(decisionBuilder));
         performOperations();
         return getCurrentStatement();
     }
 
-    private boolean isSelectForUpdate(String sql) {
-        return sql != null && (sql.endsWith("for update") || sql.endsWith("FOR UPDATE"));
-    }
-
-    private boolean isUpdate(String sql) {
-        return sql != null && (sql.startsWith("update") || sql.startsWith("UPDATE"));
-    }
-
-    private boolean isDelete(String sql) {
-        return sql != null && (sql.startsWith("delete") || sql.startsWith("DELETE"));
-    }
-
     protected Statement getWriteStatement(RouteDecisionBuilder decisionBuilder) {
         isWriteOperation = true;
+        decisionBuilder.isWrite(true);
         return prepareWriteStatement(decisionBuilder);
     }
 
--- a/src/main/java/com/atlassian/db/replica/internal/RouteDecisionBuilder.java
+++ b/src/main/java/com/atlassian/db/replica/internal/RouteDecisionBuilder.java
@@ -9,6 +9,8 @@ public final class RouteDecisionBuilder {
     private String sql = null;
     private Reason reason;
     private RouteDecision cause = null;
+    private Boolean isWrite  = null;
+
 
     public RouteDecisionBuilder(Reason reason) {
         this.reason = reason;
@@ -29,12 +31,17 @@ public final class RouteDecisionBuilder {
         return this;
     }
 
+    public RouteDecisionBuilder isWrite(final boolean isWrite) {
+        this.isWrite = isWrite;
+        return this;
+    }
+
     public String getSql() {
         return sql;
     }
 
     public RouteDecision build() {
-        return new RouteDecision(sql, reason, cause);
+        return new RouteDecision(sql, reason, cause, isWrite);
     }
 
     @Override
--- a/src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
+++ b/src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
@@ -1,0 +1,50 @@
+package com.atlassian.db.replica.internal;
+
+public class SQLOpeartionType {
+
+    public static enum SQLType {
+        WRITE, READ_ONLY, NOT_KNOWN
+    }
+
+    private final SqlFunction sqlFunction;
+
+    public SQLOpeartionType(SqlFunction sqlFunction) {
+        this.sqlFunction = sqlFunction;
+    }
+
+    /**
+     * Arbitrially judge based on sql param whether sql is WRITE
+     */
+    public boolean whetherSQLisWrite(String sql){
+        //method to arbitrially judge whether sql isWrite or
+        //maybe inspired MaryPri
+        //https://bitbucket.org/atlassian/marypri/src/00f9cad9d58131e6d9d27be8de0ee5d01b3cbfd3/src/main/kotlin/io/atlassian/micros/marypri/sql/ReadWriteClassifier.kt#lines-14
+        return isWriteOpearation(sql) &&
+                isSelectForUpdate(sql) &&
+                !isSQLSet(sql);
+    }
+
+
+    public boolean isWriteOpearation(String sql) {
+        boolean isWriteOperation = sqlFunction.isFunctionCall(sql) || isUpdate(sql) || isDelete(sql);
+
+        return isWriteOperation;
+
+    }
+
+    public boolean isSelectForUpdate(String sql) {
+        return sql != null && (sql.endsWith("for update") || sql.endsWith("FOR UPDATE"));
+    }
+
+    public boolean isSQLSet(String sql) {
+        return sql.startsWith("set");
+    }
+
+    public boolean isUpdate(String sql) {
+        return sql != null && (sql.startsWith("update") || sql.startsWith("UPDATE"));
+    }
+
+    public boolean isDelete(String sql) {
+        return sql != null && (sql.startsWith("delete") || sql.startsWith("DELETE"));
+    }
+}
--- a/src/test/java/com/atlassian/db/replica/api/reason/TestRouteDecision.java
+++ b/src/test/java/com/atlassian/db/replica/api/reason/TestRouteDecision.java
@@ -1,0 +1,44 @@
+package com.atlassian.db.replica.api.reason;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+public class TestRouteDecision {
+
+    private static final RouteDecision READ_DECISION = new RouteDecision("select * from somewhere", Reason.READ_OPERATION, null);
+    private static final RouteDecision WRITE_DECISION = new RouteDecision("select * from somewhere", Reason.WRITE_OPERATION, null);
+    private static final RouteDecision WRITE_DECISION2 = new RouteDecision("select * from somewhere", Reason.LOCK, null);
+    private static final RouteDecision WRITE_DECISION3 = new RouteDecision("select * from somewhere", Reason.RW_API_CALL, null);
+
+    @Parameters
+    public static Collection routeDecisions() {
+        return Arrays.asList(new Object[][]{
+                {READ_DECISION, false},
+                {WRITE_DECISION, true},
+                {WRITE_DECISION2, true},
+                {WRITE_DECISION3, true},
+        });
+    }
+
+    private RouteDecision givenRouteDecision;
+    private boolean expectedIsWrite;
+
+    public TestRouteDecision(RouteDecision givenRouteDecision, boolean expectedIsWrite) {
+        this.givenRouteDecision = givenRouteDecision;
+        this.expectedIsWrite = expectedIsWrite;
+    }
+
+    @Test
+    public void shouldDetectWriteDatabaseCallAndRunReleatedObservedAction() {
+        assertThat(givenRouteDecision.isWrite()).isEqualTo(expectedIsWrite);
+    }
+
+}
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaPreparedStatement.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaPreparedStatement.java
@@ -380,6 +380,7 @@ public class ReplicaPreparedStatement extends ReplicaStatement implements Prepar
         if (currentStatement != null) {
             return currentStatement.getParameterMetaData();
         } else {
+            //todo writeStatment not always is used for write SQL queries
             return getWriteStatement(new RouteDecisionBuilder(RW_API_CALL)).getParameterMetaData();
         }
     }


afaruga-atlassian:
Screenshot 2021-03-02 at 10 13 12
Github is cool. I didn't know about this feature.
Thanks @dagguh !

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -554,41 +561,35 @@ void recordWriteAfterQueryExecution() throws SQLException {
 
     public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) {
         if (connectionProvider.getState().equals(MAIN)) {
-            decisionBuilder.reason(MAIN_CONNECTION_REUSE);
+            decisionBuilder.reason(MAIN_CONNECTION_REUSE);//.isWrite(true) ? ;
+            // sth was readonly, but need to happen on Main because of inconsistency
```
dagguh, line 565: The `isWrite` check needs to happen before the `isConsistent` check

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
```diff
@@ -0,0 +1,50 @@
+package com.atlassian.db.replica.internal;
+
+public class SQLOpeartionType {
```
dagguh, line 3:
```suggestion
public class SQLOperationType {
```
There are more "opeartion " typos in the PR

wyrzyk:
> Is there a way to arbitrarily judge whether SQL is read/write?

Can we start from the context? Why do we need to know if it's a read/write operation?

Some operations may be read-only but will never reach replica. It also depends on the context. 

>  How SQL operation is treated in db-replica is based on statementType + sql + consistency

+ context/status, transaction isolation level (I'm not 100% sure if it covers all the ingredients)

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    private final SqlFunction sqlFunction;
+
+    public SQLOpeartionType(SqlFunction sqlFunction) {
+        this.sqlFunction = sqlFunction;
+    }
+
+    /**
+     * Arbitrially judge based on sql param whether sql is WRITE
```
dagguh, line 16: Typo in "Arbitrially".
Why arbitrarily though?

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public boolean whetherSQLisWrite(String sql){
```
dagguh, line 18:
```suggestion
    public boolean writesToDatabase(String sql){
```

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh, line 3: The OOP way:
```java
public class SqlQuery {

    private final String query;
...
    public Boolean writesToDatabase() {
```

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
```diff
@@ -0,0 +1,50 @@
...
+        //method to arbitrially judge whether sql isWrite or
+        //maybe inspired MaryPri
```
dagguh, line 20: MaryPri uses "is JQL", which is Jira-specific. It also uses a `org.postgresql.core.Parser`, which is Postgres-specific. Parsers are also a lot slower than String heuristics.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public boolean isSQLSet(String sql) {
+        return sql.startsWith("set");
+    }
+
+    public boolean isUpdate(String sql) {
+        return sql != null && (sql.startsWith("update") || sql.startsWith("UPDATE"));
+    }
+
+    public boolean isDelete(String sql) {
```
dagguh, line 47: Why is all of this public?

src/test/java/com/atlassian/db/replica/api/reason/TestRouteDecision.java
```diff
@@ -0,0 +1,44 @@
...
+    private RouteDecision givenRouteDecision;
+    private boolean expectedIsWrite;
+
+    public TestRouteDecision(RouteDecision givenRouteDecision, boolean expectedIsWrite) {
+        this.givenRouteDecision = givenRouteDecision;
+        this.expectedIsWrite = expectedIsWrite;
+    }
+
+    @Test
+    public void shouldDetectWriteDatabaseCallAndRunReleatedObservedAction() {
```
dagguh, line 40: `AndRunReleatedObservedAction` part is not implemented

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -554,41 +561,35 @@ void recordWriteAfterQueryExecution() throws SQLException {
...
         }
 
-        if (isSelectForUpdate(sql)) {
-            decisionBuilder.reason(LOCK);
+        if (sqlOpeartionType.isSelectForUpdate(sql)) {
+            decisionBuilder.reason(LOCK).isWrite(true);
             return prepareWriteStatement(decisionBuilder);
         }
 
+        // todo what should happen here UNKNOWN SQL operation type ?
```
dagguh, line 584: It shouldn't be unknown. The `sql` should be checked

afaruga-atlassian:
> Can we start from the context? Why do we need to know if it's a read/write operation?

Context is, that for resource owner(REST endpoint/ amq/other) we want to let them monitor their endpoints whether actually, they are readonly, and later help them maintain that readability, to give them API that they can implement a way how to react if somebody broke their resource and make it none-readonly again, thats the idea.
Thanks for this resource owner will hope that its traffic was moved to RR.

So the question is?
Whether Resource owner is interesting on fixed information based on sql whether its resource is ReadOnly, or whether its resource its treated by db-replica as treat only, what is more complex. 😕 

> Some operations may be read-only but will never reach replica. It also depends on the context.

What's the reason? Isn't permanent, or temporary(early stages of db-replica evolution), mixed?
 
I can imagine that explaining the full context might be tricky and there is still some space to discover and learn about it.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -554,41 +561,35 @@ void recordWriteAfterQueryExecution() throws SQLException {
...
+        isWriteOperation = sqlOpeartionType.isWriteOpearation(sql);
         if (isWriteOperation) {
-            decisionBuilder.reason(WRITE_OPERATION);
+            decisionBuilder.reason(WRITE_OPERATION).isWrite(true);
             return prepareWriteStatement(decisionBuilder);
         }
 
-        if (isSelectForUpdate(sql)) {
-            decisionBuilder.reason(LOCK);
+        if (sqlOpeartionType.isSelectForUpdate(sql)) {
```
dagguh, line 579: So locking is a type of a write. The `SqlQuery.writesToDatabase` should catch that. So "writing to DB" is a `Reason` to go to replica, but there are multiple types of writes. Perhaps `SqlQuery` needs to return something more than a `boolean`. Maybe it needs more explitict methods:
```java
boolean mustWriteToDb();
boolean mayWriteToDb();
String getWritingKind(); // e.g. LOCK, UPDATE, DELETE, FUNCTION
```

dagguh (changes requested):

dagguh:
> Can we start from the context? Why do we need to know if it's a read/write operation?

I described the use case in #50

wyrzyk:
> Context is, that for resource owner(REST endpoint/ amq/other) we want to let them monitor their endpoints whether actually, they are readonly, and later help them maintain that readability, to give them API that they can implement a way how to react if somebody broke their resource and make it none-readonly again, thats the idea.

It seems the resource owner wants to know if queries are landing on replica or main and not if the queries are read-only or not. The API to do that already exist, if `DatabaseCall` uses only the above reasons, then it will work with replica-only:
```
  public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT", true);
    public static final Reason READ_OPERATION = new Reason("READ_OPERATION", false);
    public static final Reason MAIN_CONNECTION_REUSE = new Reason( "MAIN_CONNECTION_REUSE",    true );
    public static final Reason RO_API_CALL = new Reason("RO_API_CALL", false);
```
REPLICA_INCONSISTENT -> it's outside of control for a single endpoint.
MAIN_CONNECTION_REUSE -> may be caused by REPLICA_INCONSISTENT

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
wyrzyk: What's the use case for revealing these low-level details in the API.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dwatl: There are libraries that are parsing SQLs. You'd need to cache results if it's slow.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dwatl: typo: SQL**Opeartion**Type

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -94,7 +95,8 @@ public void close() throws SQLException {
     @Override
     public int getMaxFieldSize() throws SQLException {
         checkClosed();
-        return getWriteStatement(new RouteDecisionBuilder(RW_API_CALL)).getMaxFieldSize();
+        //todo this looks to be READ operation but run on write statement?
```
wyrzyk, line 98: It's easier to track it with issues than todos (see https://github.com/atlassian-labs/db-replica/issues/37 for example)

src/test/java/com/atlassian/db/replica/api/reason/TestRouteDecision.java
dwatl: "And" in tests is a code smell

src/main/java/com/atlassian/db/replica/internal/RouteDecisionBuilder.java
```diff
@@ -29,12 +31,17 @@ public RouteDecisionBuilder cause(final RouteDecision cause) {
         return this;
     }
 
+    public RouteDecisionBuilder isWrite(final boolean isWrite) {
```
dwatl, line 34: isWrite -> write / setWrite?

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh: Caching won't help, we're parsing a different `sql` each time

wyrzyk: > There are libraries that are parsing SQLs.

https://github.com/atlassian-labs/db-replica/issues/9

wyrzyk: We are building better observability. If the current approach does not work well, then we can consider adding a lib. We need to take into account all the factors:
- parsing will be slower, caching may not help and will add more complexity,
- adding dependencies to a library reduces reusability.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: I'm talking about internals. We keep hearing that "LOCK" is a type of write. Here's a way to model that explicitly.

wyrzyk: Make sense.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: that depends,
I guess i should narrowed it down

- [ ]  `isWrite` -> `isSQLWrite`
- [ ]  `isWrite` -> `wasItTreatedLikeWrite` 

Then it answers whether consistency should be 1 of the factors to build it.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: I meant finally, without discussion, ultimately would be a better word here, sorry :)

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: Thanks for the suggestion.
TBH I wasn't feeling confident about creating issues, and just created `draft` PR with bunch of comments to share my concerns and get feedback whether they're relevant.

wyrzyk: No worries. The draft was a good call. It engaged multiple people from the team. I guess the final PR will benefit from this early draft. We can create the issues not to lose them or fix them in the PR (I guess it's outside the scope).

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: @dagguh 
I've checked and `isConsistent` happens at the end of this method in `setCurrentStatement` when all reasons to prepareWriteStatement have been filter out, and still Replica is in the game for given query or rather statement.

dagguh: Nice, so maybe `Reason.REPLICA_INCONSISTENT` is not hiding anything and maybe #50 can be solved with the current API?

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: No! writesToDatabase might happen for READONLY Endpoint when Replica inconsistent, db-replica has some bug / is broken. But this doesn't change the type of SQL query READ/WRITE

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: This is just an example. WIth parsing by external lib coupled with postgress. But it would be interesting to compare db-replica judgment(READ/WRITE) on Hello for 100% of queries with MaryPri and look for space for improvement

afaruga-atlassian: In db-replica case the information about sql operation type should be known in compilation lvl as it operates on Statement & SQL low lvl internals.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: It shouldn't. 
But it might be, if SQL operation type is different thing than isWriteOperation.
What do you mean by 
> sql should be checked

Should it be parsed?

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh: I'm confused. I'm just suggesting cleaner English without changing semantics.
Call site:
```java
var type = ...
if (type.writesToDatabase("UPDATE ...")) { ...
```

It's still procedural style, so OOP style might be clearer:
```java
var query = new SqlQuery("UPDATE ...")
if (query.writesToDatabase()) { ...
```

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: So it was checked already with `isWriteOpearation` and `isSelectForUpdate`. What other possible write types are there?
As @wyrzyk [said](https://github.com/atlassian-labs/db-replica/pull/48#discussion_r585359230), the library knows. The code is just arranged in a way that's hard to extract that knowledge. We can refactor internals as much as we need. Incrementality is less important than feature power.

afaruga-atlassian:
> > Context is, that for resource owner(REST endpoint/ amq/other) we want to let them monitor their endpoints whether actually, they are readonly, and later help them maintain that readability, to give them API that they can implement a way how to react if somebody broke their resource and make it none-readonly again, thats the idea.
> 

In my opinion Resource owner would like to know what it can do to make resource RR compatible
Temporary issues like an inconsistent replica, and final outcome (whether it was run on main) might be useful to

Full data (the data might change as db-replica implementation will growth and make decision more complicated) :
IsSQLWriteOpeartaion, wasItRunOnMain, wasReplica consistent for this query.

> It seems the resource owner wants to know if queries are landing on replica or main and not if the queries are read-only or not. The API to do that already exist, if `DatabaseCall` uses only the above reasons, then it will work with replica-only:
> 
> ```
>   public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT", true);
>     public static final Reason READ_OPERATION = new Reason("READ_OPERATION", false);
>     public static final Reason MAIN_CONNECTION_REUSE = new Reason( "MAIN_CONNECTION_REUSE",    true );
>     public static final Reason RO_API_CALL = new Reason("RO_API_CALL", false);
> ```
> 
> REPLICA_INCONSISTENT -> it's outside of control for a single endpoint.
> MAIN_CONNECTION_REUSE -> may be caused by REPLICA_INCONSISTENT

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: I'm lost. Can you explain?

dagguh: It seems there is no actual case of `Reason.REPLICA_INCONSISTENT` hiding a write.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: Ok. We misunderstood.  I've had different expectation from this.  You stick to the implementation, I'm still not sure about proper meaning of `isWrite`

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: Ok, I don't get yet how to extract that knowledge and I'm asking for any suggestion related to this. I've shared my try to add isWrite information. to `RouteDecision`
But first we're unsure about final definition of isWrite

https://github.com/atlassian-labs/db-replica/pull/49#issuecomment-788925839
https://github.com/atlassian-labs/db-replica/pull/49#issuecomment-788778672

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh: There are many comments. Some affect big portions of the diff. Those are under discussion for longer.
Some are small, like typos, better naming, local refactoring. Those are quickly agreed upon and can be fixed immediately.
Even if a wide change will affect the small changes, the small ones are are worth doing anyway (just to reduce the comments).

wyrzyk:
> In my opinion Resource owner would like to know what it can do to make resource RR compatible

The resource owner will get the SQL. The user can also capture a stack-trace. We can't provide more information than that. All our internal classifications will always be less accurate than this (SQL query and context are inputs).

And then there are two cases:
1. It's a write operation -> change the code in the application to avoid it or don't assume the endpoint is read-only.
2. It's a write operation -> create an issue (or PR), there's a bug in the library.

> So for REPLICA_INCONSISTENT and MAIN_CONNECTION_REUSE what information Resource owner should receive?

REPLICA_INCONSISTENT -> Consistency of the replica reduces the effectiveness of the library. The user can experiment with better consistency implementations or with losing some requirements.

MAIN_CONNECTION_REUSE -> it will always come with a reason. It means it was already decided, and the current decision doesn't matter. User can dig into the cause `getCause`.

> wasItRunOnMain, wasReplica 

It never was. `DatabaseCall` implementation controls the database call.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
wyrzyk: We should ensure the contract is covered by a test.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: I like the idea.

SqlQuery.writesToDatabase() I think to answer this not only query is needed, but also the `context`

What's the flow /source of the method call it (I mean `execute`,`executeQuery`, `executeUpdate` etc.. 
If you think about expanding it even more-> isReplicaConsisten might be required too.

This make this complex, don't you think so ?
Or maybe you have better idea to implement it ?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: If `Reason`s already have all the information, then we don't need to rewire the internals. But we should still provide new API from #48 to keep the semantics local. E.g. in case we add a new reason or we decide to change the internal contract.

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh: It's just a query, no context

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: I've pushed new changes, it looks like REPLICA_INCONSISTENT and MAIN_CONNECTION_REUSE contain information that isSQLWrite= false ( I've introduced isSQLWrite to enforce split between consistency and isWrite) for this Reasons are enough, though it wasn't obvious at the first glance)

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: I havent introduce SQLQuery, it wasnt required this time. But it's worth to remember, maybe next time there will be more reasons to push it.

afaruga-atlassian:
I've reused stateListner to learn more about states changed behind the scene. It was much more productive than debugging.
Double responsibility of RouteDecisionBuilder to choose/mark statement(read/write) and connectionState adding coupling and make it a bit more complex. We might think about splitting it in the future. 🤷

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,28 +41,45 @@ public Reason getReason() {
         return Optional.ofNullable(cause);
     }
 
+    /**
+     * @return whether sqlQuery is recognized as writeOperation
+     */
+    public Optional<Boolean> isSqlWrite() {
```
wyrzyk, line 47: `Optional<Boolean>` 🤨

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -10,11 +10,14 @@
...
     private final String sql;
     private final RouteDecision cause;
+    private final Boolean isSqlWrite;
 
-    public RouteDecision(String sql, Reason reason, RouteDecision cause) {
+    public RouteDecision(String sql, Reason reason, RouteDecision cause, Boolean isSqlWrite) {
         this.sql = sql;
         this.reason = reason;
         this.cause = cause;
+        this.isSqlWrite = isSqlWrite;
```
wyrzyk, line 19: It can be null. Does it mean it's an optional parameter? Why it's nullable? Can we refactor it to use a builder instead of a constructor?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -273,13 +274,15 @@ public void clearBatch() throws SQLException {
     public int[] executeBatch() throws SQLException {
         checkClosed();
         final RouteDecisionBuilder decisionBuilder = new RouteDecisionBuilder(RW_API_CALL);
+        //todo is it write or read?
```
wyrzyk, line 277: - [ ] Get rid of todo's before merging.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -331,6 +334,7 @@ public int executeUpdate(String sql, String[] columnNames) throws SQLException {
     @Override
     public boolean execute(String sql, int autoGeneratedKeys) throws SQLException {
         checkClosed();
+        //all execute(***) cases might be executed by readSQL, but we treat them as Write, this might be optimization for future
```
wyrzyk, line 337: Please create an issue. Our application would probably have no gain (we use mostly prepared statements and use `#execute` for locks only), but it can be a big change for other apps.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -690,7 +701,7 @@ public void shouldExecuteReadOperationOnReplica() throws SQLException {
             .containsOnly(REPLICA);
         verify(databaseCall).call(
             any(),
-            eq(new RouteDecisionBuilder(Reason.READ_OPERATION).sql(SIMPLE_QUERY).build())
+            eq(new RouteDecisionBuilder(Reason.READ_OPERATION).sql(SIMPLE_QUERY).isSqlWrite(false).build())
```
wyrzyk, line 704: Now, READ_OPERATION sqlWrite=true is possible 😢 . Why can't we keep reducing the states at the compile time?

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -752,13 +763,135 @@ public void shouldReuseMainConnection() throws SQLException {
...
+                        new RouteDecisionBuilder(Reason.RW_API_CALL).sql(SIMPLE_QUERY).isSqlWrite(true).build()
                     )
+                    .isSqlWrite(false) //read which was treated as write
                     .build()
             )
         );
     }
 
+    @Test
+    public void shouldShowThatMainConnectionCantBeReusedAsRead() throws SQLException {
```
wyrzyk, line 775: It seems the test abuses Mockito. Can you rewrite it without (or at least limit the usage) Mockito to improve readability?

wyrzyk: It seems like the test tests multiple things at once can it follow given-when-then or arrange-act-assert pattern?

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh: "Wasn't required"? I'd say `SQLOperationType` isn't required.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
dagguh: It can even be an `Optional` of `null` 😬

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -553,42 +557,31 @@ void recordWriteAfterQueryExecution() throws SQLException {
...
-        return sql != null && (sql.startsWith("update") || sql.startsWith("UPDATE"));
-    }
-
-    private boolean isDelete(String sql) {
-        return sql != null && (sql.startsWith("delete") || sql.startsWith("DELETE"));
-    }
-
     protected Statement getWriteStatement(RouteDecisionBuilder decisionBuilder) {
         isWriteOperation = true;
+        decisionBuilder.isSqlWrite(true);
```
dagguh, line 584: Here `String sql` isn't checked, which breaks the behavioral contract of `@return whether sqlQuery is recognized as writeOperation`

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: https://github.com/atlassian-labs/db-replica/issues/53

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: True,` SQLOperationType` isn't required too. But i think it's natural extraction of all methods with similar responsibility to separate class.

afaruga-atlassian: As with `SqlQuery` this brings more OOP while currently `String sql` is used widely in whole db-replica. 
I don't see a reason to extract it to a separate class everyhwere, especially as `isWrite` methods are used only inside `ReplicaStatement` and in their context. 
Why in their context?
We can say whether SqlQuery is fully read, as this is recognized on ReplicaStatement, and depends from which method was called.

Maybe you had different idea for the use of `SQL query` ?  I'm open for suggestion.

afaruga-atlassian: I've changed `SQLOperationType` it od  `SqlFunctionChecker`

wyrzyk: `*Checker`... 🤔 . Will it trigger @dagguh? 🍿  😄

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: Shame on me!

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
dagguh: I don't understand the defense of procedural programming.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
afaruga-atlassian: I thought if `Optional<String> getSql()` then it make sense to have isSqlWrite Optional too.
I'll  left it as nullable Boolean, as it's nullable. 

> Can we refactor it to use a builder instead of a constructor?
I don't see how this might help with nullAbility.
Builder will pass null and all will work.

wyrzyk: > I don't see how this might help with nullAbility.

It will not help with nullability. It will help with usability and [API evolution](https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#3612-consider-a-builder-when-faced-with-many-constructor-parameters).

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: @wyrzyk Can you explain ?
READ_OPERATION -> means that always sqlWrite=false.
 
I'd like to reduce state at the compile time. How I can achieve that?

wyrzyk: What will happen in case I use the API this way?
```
new RouteDecisionBuilder(Reason.READ_OPERATION).sql(SIMPLE_QUERY).isSqlWrite(true).build()
```

afaruga-atlassian: aaaa I think I know what you meant!
I've tried hard to show that Reasons doesn't reflect unambiguously isWrite, but I failed and looked like they reflect, at least considering what we know right know and that we're aware of https://github.com/atlassian-labs/db-replica/issues/53
So actually I can continue with 1st idea to add isWrite to Reason!

`    private final String name;
    private final boolean isRunOnMain;
    private final boolean isWrite;
    

    private Reason(final String name, boolean isRunOnMain, boolean isWrite) {
        this.name = name;
        this.isRunOnMain = isRunOnMain;
        this.isWrite = isWrite;
    }

    public static final Reason RW_API_CALL = new Reason("RW_API_CALL", true, true);
    public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT", true, false);
    public static final Reason READ_OPERATION = new Reason("READ_OPERATION", false, false);
    public static final Reason WRITE_OPERATION = new Reason("WRITE_OPERATION", true, true);
    public static final Reason LOCK = new Reason("LOCK", true, true);
    public static final Reason MAIN_CONNECTION_REUSE = new Reason(
        "MAIN_CONNECTION_REUSE",
        true,
        false);
    public static final Reason HIGH_TRANSACTION_ISOLATION_LEVEL = new Reason(
        "HIGH_TRANSACTION_ISOLATION_LEVEL",
        true,
        false);
    public static final Reason RO_API_CALL = new Reason("RO_API_CALL", false, false);
`

wyrzyk: > I'd like to reduce state at the compile time. How I can achieve that?

In the same way as `isRunOnMain` it would have to be mapped 1:1 with reasons. We may need a second class to map it.

afaruga-atlassian: Or 2nd idea
IsWrite()
reason.equals ( RW_API_CALL | WRITE_OPERATION | LOCK ) 

 :) 
this might be simpler but it left ambiguity in the code for further devs.
Thats why I'm for 1st option.

wyrzyk: OTOH, Maciej wanted to bind query with `writesToDatabase`, this way we can't make it compile time as queries are not known at this time 🤔 .

wyrzyk: If not possible to make it OO and compile-time safe, we can introduce runtime checks.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: Ha! Good catch. That's true! Cause `db-replica` recognizes whether query was write/read based on the method that call it. (based on `Statement` API) 
It works for almost all cases except from execute, which is mentioned here https://github.com/atlassian-labs/db-replica/issues/53 and It works for us as currently as our only product currently doesn't used execute() with read queries. Anyway this will be solved and isWrite will be set on compile time
https://github.com/atlassian-labs/db-replica/pull/49#discussion_r588348234

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: I like both approaches.

dagguh: A `Reason` (or a whole `RouteDecision`) can be inferred **from** `SqlQuery` rather than independently. OO✔️ compile-time✔️

afaruga-atlassian: Good discusion! 👍 

> A Reason (or a whole RouteDecision) can be inferred from SqlQuery 
Hmm but where is the information whether Sql was invoked as writeStatement or readStatement which is pre requisite to determine `isWrite` or Reason itself.

I think I can make it OOP and compile time but SQLQuery would have to have information about read/write statement as we don't parse SQL itself and this is our source of knowledge.
Or create SqlQuery with boolean .

@dagguh  How you see it both OO and compile-time ?

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: > defense of procedural programming

Ok those words convinced me.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: I'll go with

```
public class SqlQuery {

    private final String query;
    private final boolean isWriteStatement;
    
    public SqlQuery(String query, boolean isWriteStatement) {
        this.query = query;
        this.isWriteStatement = isWriteStatement;
    }

    public Boolean writesToDatabase() {
```

afaruga-atlassian: Give me 15 mins and there will be PR, I think you will be happy with it.

dagguh: `SqlQuery` doesn't need to know about "the context" like `boolean isWriteStatement`. It's just a query.
I said that **a** (some) reason can be inferred, rather than **all** reasons.

dagguh: Concretely we could ask `SqlQuery` for `RouteDecision route()`.

afaruga-atlassian: The problem is that over there RouteDecision isn't known, and rather it's established overthere, what worse, reason can change later, down in the code in when ConnectionState is determined as RouteDecision is derivative from SQL + statement + connection state (and order of those checks too).

RouteDecision is built just before doing execute
```
return execute(statement::execute, decisionBuilder.build());
```


I'll try to somehow make it OOP, and compileTime + runtime. Either SqlQuery need to have SqlQueryBuilder, and rather RouteDecision will contain SqlQuery (as it currently contain sql) .

I'm not sure whether SqlQuery should be added to the API 🤔 

I like your thinking @dagguh, but looks like you missed some of impl detail which blocks the implementation of it.
once again  RouteDecision is derivative from SQL + statement + connection state

Another broken part is that SQL function maybe customized: ``SqlFunction(Set<String> readOnlyFunctions)`` and this need to come from DualConnectionBuilder.
That's why SQLQuery  goes together with Reason, not independently

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: Actually `LOCK` isn't type o write, but it's treat almost as write as it run for on main. acutall. `LOCK` == `SELECT FOR UPDATE`
I added isWrite information to Rason on compile level , this way this model say that explicitly. Thanks for this I could remove `isWriteOpearatiion` internal state, before doing this I've improved code coverage for `SELECT FOR UPDATE`.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: I've pushed changes wtih compile time explicit isWrite information on Reason lvl. I also introduce SqlQuery to increase OO. 
It's not exactly the same because of mentioned reasons(mainly `RouteDecision` is build later, and `RouteDecisionBuilder` is used not only for Statements(SQL) but for choosing connection(based on connection state) - in those places adding `SqlQuery` doesn't make sense.

This might be improved in future but for now it's +1 improvement into good direction.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: done

src/main/java/com/atlassian/db/replica/internal/SQLOpeartionType.java
afaruga-atlassian: `SqlQuery` introduced, Friday afternoon wasn't productive, class name with verbs aren't good idea. 🤦

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -8,26 +8,28 @@
...
         this.isRunOnMain = isRunOnMain;
+        this.isWrite = isWrite;
     }
 
-    public static final Reason RW_API_CALL = new Reason("RW_API_CALL", true);
-    public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT", true);
-    public static final Reason READ_OPERATION = new Reason("READ_OPERATION", false);
-    public static final Reason WRITE_OPERATION = new Reason("WRITE_OPERATION", true);
-    public static final Reason LOCK = new Reason("LOCK", true);
+    public static final Reason RW_API_CALL = new Reason("RW_API_CALL", true, true);
```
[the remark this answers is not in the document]
afaruga-atlassian: No. they're independent. This is not code smell in this place(constructor args)

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -8,26 +8,28 @@
...
+    public static final Reason REPLICA_INCONSISTENT = new Reason("REPLICA_INCONSISTENT", true, false);
+    public static final Reason READ_OPERATION = new Reason("READ_OPERATION", false, false);
+    public static final Reason WRITE_OPERATION = new Reason("WRITE_OPERATION", true, true);
+    public static final Reason LOCK = new Reason("LOCK", true, false);
     public static final Reason MAIN_CONNECTION_REUSE = new Reason(
         "MAIN_CONNECTION_REUSE",
-        true
-    );
+        true,
```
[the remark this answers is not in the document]
afaruga-atlassian: At the end turns out yes!!!  Reason can point out SQL write queries. Though it might happen that not write will still be run on Main due to connection state/ transaction. The reason is derivative from couple of variables (statement,SQL, connection) and it used for couple of reason to build :

-  read/write statement, 
-  read/write connection
And at the same time RouteDecision is used to determine statement/connection(state) 

In future new combinations of SQL /statement/connection should end up building new Reasons.
e.g. now we have LOCK => only for `SELECT FOR UPDATE` ( we may have more like this)

It's much more explicit right know on compile lvl

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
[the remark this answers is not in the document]
wyrzyk: @afaruga-atlassian, I think @dwatl wanted to say it's hard to read and understand. It's also easy to make an error and put the arguments in the wrong order. Kotlin solves this problem with named arguments. In Java, we can use a builder for `Reason` to make it more readable.

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -57,9 +73,11 @@ public int hashCode() {
     @Override
     public String toString() {
         return "RouteDecision{" +
-            "reason=" + reason +
-            ", sql='" + sql + '\'' +
-            ", cause=" + cause +
-            '}';
+                "reason=" + reason +
```
afaruga-atlassian, line 76: Looks like I have the wrong IntelliJ setup:(

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,15 +38,31 @@ public Reason getReason() {
...
+    /**
+     * @return Decision whether current route is Run on Main or Replica
+     */
+    public boolean isRunOnMain() {
+        return reason.isRunOnMain();
+    }
+
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
```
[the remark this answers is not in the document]
afaruga-atlassian: sure, wrong intellij setup, thanks!

wyrzyk (changes requested):
The change can't be rolled-out without disabling DualConnection. I created PRs for two propositions on how to roll out changes safely (https://github.com/atlassian-labs/db-replica/pull/52 and https://github.com/atlassian-labs/db-replica/pull/40). We can choose one of them or a different approach, but we shouldn't release it without any safe-rollout mechanism.

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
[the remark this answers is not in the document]
dwatl: Yes, let's imagine a function `f(true, false false, true)`. It's easy to shoot your own foot with it.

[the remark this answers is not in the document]
afaruga-atlassian: done, good suggestion 👍

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
dagguh: Just try it

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -0,0 +1,30 @@
+package com.atlassian.db.replica.internal;
+
+public class SqlQuery {
+
+    private final String sql;
+
+    public SqlQuery(String sql) {
```
dagguh, line 7: Thanks!

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: If it's not a write, then why is it run on main?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -24,7 +24,7 @@
     private final List<StatementOperation<Statement>> batches = new ArrayList<>();
     private final ReplicaConsistency consistency;
     private final DatabaseCall databaseCall;
-    private boolean isWriteOperation = true;
```
afaruga-atlassian, line 27: @wyrzyk 
> The change can't be rolled-out without disabling DualConnection. I created PRs for two propositions on how to roll out changes safely (#52 and #40). We can choose one of them or a different approach, but we shouldn't release it without any safe-rollout mechanism.

Oh man! This is scope explosion. Though I understand your intentions. It's just those changes aren't merged yet.
Just to double check,
what's the risk of rolling out `DualConnection` with this change from yours perspective? From my point the only risky is removing `isWriteOperation`  internal field, which is proven by unit tests, what give me confidence on compile lvl to go 👍 

Maybe I can extract this change and then merge It once https://github.com/atlassian-labs/db-replica/pull/52 will be merged?

wyrzyk: > what's the risk of rolling out DualConnection with this change from yours perspective? 

We need to use FF for every rollout (deployment may run outside of our working hours and we are expected to observe the rollout). We also should stop to use global kill-switch. So far I have two propositions about how to solve this problem, I'm also open to other solutions.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: I guess this is related with the impact on transaction
> FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update. This prevents them from being modified or deleted by other transactions until the current transaction ends.
[docs](https://www.postgresql.org/docs/9.0/sql-select.html#SQL-FOR-UPDATE-SHARE)

@wyrzyk should provide full answer on this.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: Tried, I hope it met your expectations?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: Fair enough..,
So we need to prioritize (#52 and #40) 🙏 
Let me jump on reviewing (#52 and #40) to push it further as it blocks this story 🚫

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
wyrzyk: Are other connections affected by the lock? If it's not a write operation, why are they affected?

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -37,12 +41,16 @@ public boolean isRunOnMain() {
...
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Reason reason = (Reason) o;
-        return isRunOnMain == reason.isRunOnMain && Objects.equals(name, reason.name);
+        return isRunOnMain == reason.isRunOnMain && isWrite == reason.isWrite && Objects.equals(name, reason.name);
     }
 
     @Override
```
wyrzyk, line 56: `hashCode` ignores isWrite

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -55,6 +63,35 @@ public String toString() {
...
+        private String name;
+        private boolean isRunOnMain;
+        private boolean isWrite;
+
+        private ReasonBuilder(final String name) {
+            this.name = name;
+        }
+
+        static ReasonBuilder name(final String name) {
+            return new ReasonBuilder(name);
```
wyrzyk, line 80: Setting name overrides other fields!

ReasonBulder("a").name("b").isWrite(true) != ReasonBulder("a").isWrite(true).name("b")

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -55,6 +63,35 @@ public String toString() {
         return "Reason{" +
             "name='" + name + '\'' +
             ", isRunOnMain=" + isRunOnMain +
+            ", isWrite=" + isWrite +
             '}';
     }
+
+    private static class ReasonBuilder {
+        private String name;
+        private boolean isRunOnMain;
```
wyrzyk, line 72: Can you make the default values explicit?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
dagguh: If we send `SELECT FOR UPDATE` to replica - can it succeed?

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
dagguh: Do you have it on a side branch?

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
wyrzyk: I see it's static, so the name can't be changed. Let's make it `final`.

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -0,0 +1,30 @@
+package com.atlassian.db.replica.internal;
+
+public class SqlQuery {
```
wyrzyk, line 3: [ Consider declaring the class as final](https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#2314-consider-declaring-the-class-as-final)

wyrzyk: Ahh, it's internal ❤️ .

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
afaruga-atlassian: > return isRunOnMain == reason.isRunOnMain && isWrite == reason.isWrite && Objects.equals(name, reason.name);
I've added isWrite to hashCode today morning,
Do you meant sth else?

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
afaruga-atlassian: fixed

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
afaruga-atlassian: fixed

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: Do we treat connections/calls withing the same connection to  Database independently?
I think not, we had this hidden contract that read after write is treated RUN ON MAIN.
I think with `SELECT FOR UPDATE` the idea was the same.

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
afaruga-atlassian: Yes, but still I added it

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
dagguh: Each `Reason.isWrite` is statically predetermined now, so inferring it from `SqlQuery` is no longer needed for compile-time invariant 👍

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: * Separate branch for refactor https://github.com/atlassian-labs/db-replica/pull/54
* This branch will left only with changes required for adding isWrite API

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: I'll resove conversation and move to: https://github.com/atlassian-labs/db-replica/pull/55#discussion_r590029250

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -37,12 +41,16 @@ public boolean isRunOnMain() {
         return isRunOnMain;
     }
 
+    public boolean isWrite() {
```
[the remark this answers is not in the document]
dagguh: They should be hidden in `Reason`

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,6 +38,20 @@ public Reason getReason() {
...
+     * @return True if the database call would always fail on a read replica.
+     */
+    public boolean isWrite() {
+        return reason.isWrite();
+    }
+
+    /**
+     * @return Decision whether current route is Run on Main or Replica
+     */
+    public boolean isRunOnMain() {
```
dagguh, line 51: "Current route" is not well defined.
```suggestion
    /**
     * @return true if the accompanying {@link SqlCall} will be run on the main database.
     */
    public boolean willRunOnMain() {
```
`SqlCall` has to be imported or fully-qualified for proper linking.

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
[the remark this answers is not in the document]
afaruga-atlassian: done

src/main/java/com/atlassian/db/replica/api/reason/RouteDecision.java
```diff
@@ -38,6 +38,20 @@ public Reason getReason() {
         return Optional.ofNullable(cause);
     }
 
+    /**
+     * @return True if the database call would always fail on a read replica.
+     */
+    public boolean isWrite() {
+        return reason.isWrite();
+    }
```
dagguh, line 46:
```suggestion
    /**
     * @return false if the accompanying {@link SqlCall#call()} would fail when run on replica.
     */
    public boolean couldRunOnReplica() {
        return !reason.isWrite();
    }
```
Or maybe:
```suggestion
    /**
     * @return true if the accompanying {@link SqlCall#call()} would fail when run on replica.
     */
    public boolean mustRunOnMain() {
        return reason.isWrite();
    }
```

I'm trying to use symmetrical language across the `RouteDecision` API.

dagguh (dismissed):

dagguh (approved):

afaruga-atlassian:
@wyrzyk because you requested change requested I need your response to move forward with PR

wyrzyk (approved):

PR #51, opened 2021-03-02
Add high-level overview to the docs

wyrzyk:
We already have a low-level overview of `DualConnection`, but I think
we lack something user-level, to build a basic mental model.


dagguh (approved):
Respect

PR #52, opened 2021-03-04
Issue/37 use replica for native sql second approach

wyrzyk:


src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -486,6 +520,7 @@ public static Builder builder(
         private CircuitBreaker circuitBreaker = new BreakOnNotSupportedOperations();
         private StateListener stateListener = new NoOpStateListener();
         private Set<String> readOnlyFunctions = new HashSet<>();
+        private boolean compatibleWithPreviousVersion = false;
```
dagguh, line 523: So like a "safe mode", which is disabled by default.
So a consumer makes a bump and sets it to `true`. Then they experiment with `false` and if that's good, then they can think about another bump.

So all experiments are coupled together in a given version. And we cannot exclude them from compatibility mode until consumers prove they work. So the release cycle becomes dependent on the consumers.

Not saying that's a blocker.

wyrzyk: > So the release cycle becomes dependent on the consumers?

If there's a bug in some version, then the customer can keep using the previous version until fixed.

dagguh: Ok, so each release we search and destroy the current `if (compatibleWithPreviousVersion)`?

wyrzyk: Yes. I think we should add a post-release checklist:
- mark release
- bump the version in the readme
- remove all compatibility checks

wyrzyk: It forces FFs to be short living. I think it's a good thing.

dagguh: Yep and it doesn't accumulate flags as API, so they can be removed without major release

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -28,18 +54,21 @@
...
         StateListener stateListener,
-        Set<String> readOnlyFunctions
+        Set<String> readOnlyFunctions,
+        boolean compatibleWithPreviousVersion
     ) {
         this.connectionProvider = new ReplicaConnectionProvider(connectionProvider, consistency, stateListener);
         this.consistency = consistency;
         this.databaseCall = databaseCall;
         this.readOnlyFunctions = readOnlyFunctions;
+        this.compatibleWithPreviousVersion = compatibleWithPreviousVersion;
```
dagguh, line 71: So this is bound to the lifetime of a `DualConnection`. [`FeatureDecisions`](https://github.com/atlassian-labs/db-replica/pull/40#discussion_r581972605) could have been checked on every interaction.
It's a disadvantage for this solution, but doesn't look like a blocker.

dagguh (approved):

src/main/java/com/atlassian/db/replica/api/DualConnection.java
afaruga-atlassian: It force limiting WIP bugs, which is good-> LEAN approach, you rollout only 1 change and even if sth is broken you know what changes cause it.

I think about situation when worked 1,2 started in pararel. 1 stream introduced bug, but was first, 2nd stream was slower to implement but it has higher priority.   What's then?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -520,14 +555,28 @@ private Builder(
             return this;
         }
 
+        /**
+         * Puts this connection in compatibility mode with a previous version. Developers can use this method to
+         * roll out the new version of the library with a feature flag.
+         *
+         * It's best-effort, and there's no guarantee the library in compatibility mode will always behave
+         * the same way as the previous version of the library.
```
afaruga-atlassian, line 563: > here's no guarantee the library in compatibility mode will always behave
>          * the same way as the previous version of the library

Why? Because still there is room for bugs?
I guess all `compatibleWithPreviousVersion` changes have to be covered in unit tests.
I'm missing example.

What we can do to guarantee compatibility mode working the same way as the previous version ? 
Maybe we should provide 'template' with supplier before if, after if, AKA rolloutService?

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -190,15 +190,30 @@ public void shouldAvoidFetchingReadConnectionWhenNotNecessary() throws SQLExcept
...
+            .containsExactly(REPLICA);
+    }
+
+    @Test
+    public void shouldBePossibleToDisableNativeSqlFix() throws SQLException {
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).compatibleWithPreviousVersion()
```
afaruga-atlassian, line 212: Looking from customer perspective, isn't this easier to introduce a bug for a person not familiar with `safe mode`? 
So some dev will just bump the version. How we can force him to think about possibility to introduce bugs?

One way would be to force `dual-connection` consumer to define : 
`compatibleWithPreviousVersion()` or `inompatibleWithPreviousVersion()`

afaruga-atlassian (changes requested):

src/main/java/com/atlassian/db/replica/api/DualConnection.java
afaruga-atlassian: Let me try answer myself:

0. we have version 1.0.0
1st changes rolled out with 1.0.1 contain bug.
2nd higher priority changes have to be rolled out as 1.0.2 . 
So maybe `compatibleWithPreviousVersion ` should apply to major/minor changes?
MAJOR.MINOR.PATCH?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -520,14 +555,28 @@ private Builder(
...
+         *
+         */
+        public DualConnection.Builder compatibleWithPreviousVersion() {
```
afaruga-atlassian, line 566: lack of passing param force consumer to define `DualConnection.Builder` in two places, instead of just passing the boolean value from FeatureFlag. Is it intentional?  If yes why?
So much duplicated code, maybe we could simplify it?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
afaruga-atlassian: It's sooo tempting, I have to ask:
why not sth like `Map<String,Boolean>`with false return as a default. ? 
It solves problem of multiple introduced bugs, and slower development cycle because of dependency between them, but you will state that at the same time it introduce complexity.
Usually devs and good practices maintain complexity but use `feature power` to deliver faster.
WDYT?


aah shit, I know, I saw it somewhere:
https://github.com/atlassian-labs/db-replica/pull/40
Ok, I more fun of `Experiment API` plus I would couple it with semantic versions, and have an automated way to disable any Experiment in major releases.

wyrzyk: `Experiment AP`I introduces a broad API that will only grow and accumulate complexity. Removing API means major release, which is a pain for the library's consumers. Compatibility mode puts more constraints on the libraries maintenance but will not force us to choose between growing complexity and breaking compatibility.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
wyrzyk: > What we can do to guarantee compatibility mode working the same way as the previous version ?

I guess it's not possible to give such a guarantee. It's better to say it explicit than to pretend we can make backwards-compatible every change.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: @afaruga-atlassian Please keep in mind it's an open-source library. Not all the products use feature flags, and not all the products are interested in the feature. By forcing people to use it, we make the library more complicated to use. `inompatibleWithPreviousVersion` is what every library do, and it's a good default.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
wyrzyk: You don't have to define the builder in two places.
```
 final DualConnection.Builder builder = DualConnection.builder(
            connectionProvider,
            consistency
        );
        if(oldVersionFlag){
            builder. compatibleWithPreviousVersion()
        }
```
When we pass a parameter, we can expect 3 values (true, false, null). You have 3 values to handle instead of one. False has no sense as it's a default. I don't know how to handle null properly here.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
afaruga-atlassian: I think it's faster to go and try this solution here . It is simple to maintain and remove/replace if **it will impact our velocity with maintaining db-replica**, then we should go into `FeatureDecisions` and `Experiments` API

afaruga-atlassian (approved):

afaruga-atlassian:
> enabled auto-merge (rebase)

👍  Cool GH feature

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dagguh: What's a problem case?

PR #55, opened 2021-03-09
Issue/jces 1896 is write api sql refactor

afaruga-atlassian:
This PR contains 2 commits, and it was created to split feature & refactor changes, it's dependent on:
https://github.com/atlassian-labs/db-replica/pull/49/files

it should be rollout safely due to: https://github.com/atlassian-labs/db-replica/pull/49#discussion_r589308152
 
Focus only on 2nd commit: https://github.com/atlassian-labs/db-replica/pull/55/commits/e8c45c7575d9d550daeb08450bde267224c7b5f6

### DIFF
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
@@ -24,7 +24,7 @@ public class ReplicaStatement implements Statement {
     private final List<StatementOperation<Statement>> batches = new ArrayList<>();
     private final ReplicaConsistency consistency;
     private final DatabaseCall databaseCall;
-    private boolean isWriteOperation = true;
+
     private final SqlFunction sqlFunction;
     private final DecisionAwareReference<Statement> readStatement = new DecisionAwareReference<Statement>() {
         @Override
@@ -176,7 +176,8 @@ public class ReplicaStatement implements Statement {
         checkClosed();
         final RouteDecisionBuilder decisionBuilder;
         final Statement statement;
-        if (sql.startsWith("set")) {
+        SqlQuery sqlQuery = new SqlQuery(sql);
+        if (sqlQuery.isSqlSet()) {
             decisionBuilder = new RouteDecisionBuilder(READ_OPERATION).sql(sql);
             statement = getReadStatement(decisionBuilder);
         } else {
@@ -501,7 +502,7 @@ public class ReplicaStatement implements Statement {
 
     <T> T execute(final SqlCall<T> call, final RouteDecision routeDecision) throws SQLException {
         final T result = databaseCall.call(call, routeDecision);
-        if (routeDecision.getReason().isRunOnMain() && isWriteOperation) {
+        if (routeDecision.getReason().isWrite()) {
             recordWriteAfterQueryExecution();
         }
         return result;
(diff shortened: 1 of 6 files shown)


src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -24,7 +24,7 @@
     private final List<StatementOperation<Statement>> batches = new ArrayList<>();
     private final ReplicaConsistency consistency;
     private final DatabaseCall databaseCall;
-    private boolean isWriteOperation = true;
```
afaruga-atlassian, line 27: https://github.com/atlassian-labs/db-replica/pull/49#discussion_r589301633

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -558,37 +559,22 @@ public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) {
...
-        isWriteOperation = sqlFunction.isFunctionCall(sql) || isUpdate(sql) || isDelete(sql);
-        if (isWriteOperation) {
+        SqlQuery sqlQuery = new SqlQuery(decisionBuilder.getSql());
+        if (sqlQuery.isWriteOperation(sqlFunction)) {
             decisionBuilder.reason(WRITE_OPERATION);
             return prepareWriteStatement(decisionBuilder);
         }
 
-        if (isSelectForUpdate(sql)) {
+        if (sqlQuery.isSelectForUpdate()) {
```
afaruga-atlassian, line 568: Previous discussion: https://github.com/atlassian-labs/db-replica/pull/49#discussion_r585406595

afaruga-atlassian: @dagguh
So locking is a type of a write. The SqlQuery.writesToDatabase should catch that. So "writing to DB" is a Reason to go to replica, but there are multiple types of writes. Perhaps SqlQuery needs to return something more than a boolean. Maybe it needs more explitict methods:

boolean mustWriteToDb();
boolean mayWriteToDb();
String getWritingKind(); // e.g. LOCK, UPDATE, DELETE, FUNCTION
 
@wyrzyk wyrzyk 7 days ago Member
What's the use case for revealing these low-level details in the API.

 
@dagguh dagguh 7 days ago Member
I'm talking about internals. We keep hearing that "LOCK" is a type of write. Here's a way to model that explicitly.

 
@wyrzyk wyrzyk 7 days ago Member
Make sense.

 
@afaruga-atlassian afaruga-atlassian yesterday Author Member
Actually LOCK isn't type o write, but it's treat almost as write as it run for on main. acutall. LOCK == SELECT FOR UPDATE
I added isWrite information to Rason on compile level , this way this model say that explicitly. Thanks for this I could remove isWriteOpearatiion internal state, before doing this I've improved code coverage for SELECT FOR UPDATE.

 
@dagguh dagguh 22 hours ago Member
If it's not a write, then why is it run on main?

 
@afaruga-atlassian afaruga-atlassian 21 hours ago Author Member
I guess this is related with the impact on transaction

FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update. This prevents them from being modified or deleted by other transactions until the current transaction ends.
docs

@wyrzyk should provide full answer on this.

 
@wyrzyk wyrzyk 21 hours ago Member
Are other connections affected by the lock? If it's not a write operation, why are they affected?

 
@dagguh dagguh 21 hours ago Member
If we send SELECT FOR UPDATE to replica - can it succeed?

 
@afaruga-atlassian afaruga-atlassian 20 hours ago Author Member
Do we treat connections/calls withing the same connection to Database independently?
I think not, we had this hidden contract that read after write is treated RUN ON MAIN.
I think with SELECT FOR UPDATE the idea was the same.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
afaruga-atlassian: @wyrzyk 21 hours ago Member
what's the risk of rolling out DualConnection with this change from yours perspective?

We need to use FF for every rollout (deployment may run outside of our working hours and we are expected to observe the rollout). We also should stop to use global kill-switch. So far I have two propositions about how to solve this problem, I'm also open to other solutions.

wyrzyk:
@afaruga-atlassian, I feel confused. We have two open PR's for the same change, and some discussions moved. I don't know which PR to review and the value of the split (both commits are in the same PR, so both will be merged at once and released at once).

afaruga-atlassian:
> @afaruga-atlassian, I feel confused. We have two open PR's for the same change, and some discussions moved. I don't know which PR to review and the value of the split (both commits are in the same PR, so both will be merged at once and released at once).


The intention was to rollout it safely.
PR https://github.com/atlassian-labs/db-replica/pull/49 has higher priority, I think its obvious from the commits.

As safe rollout was merged,https://github.com/atlassian-labs/db-replica/pull/52 I'll use it to rollout safely those changes together.

PR #56, opened 2021-03-09
Mark release 0.1.30

wyrzyk:


src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -104,13 +104,8 @@ public CallableStatement prepareCall(String sql) throws SQLException {
     @Override
     public String nativeSQL(String sql) throws SQLException {
         checkClosed();
-        if (compatibleWithPreviousVersion) {
```
dagguh, line 107: Let's add it to `docs/releasing.md`

dagguh (approved):

PR #58, opened 2021-03-11
Keep state and state listener internal

wyrzyk:
Currently, there is no use case beside the tests. I'd keep it internal before
1.0.0 release. We can either move it to the API or remove it later.


src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -20,11 +18,11 @@
 import static com.atlassian.db.replica.api.reason.Reason.REPLICA_INCONSISTENT;
 import static com.atlassian.db.replica.api.reason.Reason.RO_API_CALL;
 import static com.atlassian.db.replica.api.reason.Reason.RW_API_CALL;
-import static com.atlassian.db.replica.api.state.State.CLOSED;
-import static com.atlassian.db.replica.api.state.State.COMMITED_MAIN;
-import static com.atlassian.db.replica.api.state.State.MAIN;
-import static com.atlassian.db.replica.api.state.State.NOT_INITIALISED;
-import static com.atlassian.db.replica.api.state.State.REPLICA;
+import static com.atlassian.db.replica.internal.state.State.CLOSED;
```
afaruga-atlassian, line 21: Can you describe the reasoning behind hiding  State.
It can't be concluded from the changes.

afaruga-atlassian:
I wonder whether this shouldn't be blocked till https://github.com/atlassian-labs/db-replica/pull/57 will be merged?
So quick develoop would stick to concumer-devloop-cycle

If not then changes would accumulate and if 0.1.30 will be broken and 0.1.31 would be merged before those problems revealed, then 0.1.31 will have those bugs, and revert will have to apply for the booth of them. This introduces some juggling between versions, what will be overhead.

Do I get it correctly ? 
If yes it would be great to add some automation and block releasing new versions till the previous ones are not confirmed as compatible.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
dagguh: Removed API goes thru a deprecation cycle. And deprecation should point to the replacement.
With SemVer 0.x.y we can break stuff all the time, but we should use deprecation anyway, because:
* the consumers will know how to bump
* it's self-explanatory for other maintainers
* we're on prod already, so [we should already be 1.0.0+](https://semver.org/#how-do-i-know-when-to-release-100)
* deprecation is an important muscle to train

src/main/java/com/atlassian/db/replica/internal/state/State.java
```diff
@@ -1,4 +1,4 @@
-package com.atlassian.db.replica.api.state;
```
dagguh, line 1: Let's mention it in the changelog

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: I'm preparing for `1.0.0` release. There's no clear use-case or user for this API because it was replaced by `RouteDecision` API (the same goal, but a different approach). It's safe to keep it out, as it gives us more options for the evolution of the API (we can bring it back, remove, or add it in a better shape after we have a real user and use-case).

dagguh:
> If yes it would be great to add some automation and block releasing new versions till the previous ones are not confirmed as compatible.

😔
So `db-replica` would have to slow down to the pace of the consumer. It's not a great dependency to have

wyrzyk:
>Removed API goes thru a deprecation cycle. And deprecation should point to the replacement.
With SemVer 0.x.y we can break stuff all the time, but we should use deprecation anyway, because:
the consumers will know how to bump
it's self-explanatory for other maintainers
we're on prod already, so we should already be 1.0.0+
deprecation is an important muscle to train

I don't understand how deprecation will help here. We are removing it from API without providing any alternative. The goal is to unlock `1.0.0` end to release without doubtful or broken features.

wyrzyk:
> So db-replica would have to slow down to the pace of the consumer. It's not a great dependency to have

I see you tend to see how the current solution slows us down but not appreciate the fact it slows us far less than keeping all feature flags between major releases. I don't even want to imagine the nightmare we have avoided.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
dagguh: > I don't understand how deprecation will help here. We are removing it from API without providing any alternative. The goal is to unlock 1.0.0 end to release without doubtful or broken features.

The alternative is `RouteDecision`.
I am all in favor of removing doubtful features. But we already got consumers. We can deprecate now and remove in the next release.
Do you know that there are no usages of `State`? Having such extra external knowledge would explain why you see deprecation as unnecessary.

wyrzyk: > We can deprecate now and remove in the next release.

We can but is it worth it?
1. Deprecate
2. Release
3. Safe-rollout
4. Remove
5. Release
6. Safe-rollout

We could put this extra effort into making the lib better instead of for formality with non-semver version when the only known customer doesn't use it and have never used it.

wyrzyk: In SemVer, there will be no extra overhead. You add feature and deprecate OLD API at the same release.

dagguh: > 3. Safe-rollout
> 5. Release
> 6. Safe-rollout

We can skip these steps.

Could you address this question?
> Do you know that there are no usages of `State`?

Or maybe deprecation isn't worth it, because we know only one consumer and we can use IDE to proactively find the usages?

We can skip the deprecation, but we need a good reason for that. I'm digging, because I care about the decision-making, not the outcome. The end result will be the same and we will move on, but the reasoning we used here will be reused later on many other occasions. It has to be solid.

Here's my reason for skipping the deprecation: [SemVer rule 4](https://semver.org/#spec-item-4). I argued that we *should be* on 1.0.0 due to [a SemVer hint](https://semver.org/#how-do-i-know-when-to-release-100), but we *are not*. So the hint is weaker than the rule.

dagguh (changes requested):
[Deprecation is not required](https://semver.org/#how-do-i-know-when-to-release-100), but we should [keep the changelog](https://github.com/atlassian-labs/db-replica/pull/58/files#r592245305).

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: > We can skip these steps.

How? 

> Do you know that there are no usages of State?

No. but It's pre `1.0.0`. It can be removed anytime.

dagguh:
> I see you tend to see how the current solution slows us down but not appreciate the fact it slows us far less than keeping all feature flags between major releases.

Sorry for that. I do appreciate it. I should express it more often.
We don't see the nested ifs and we don't have to test permutations of all the flags. That's an amazing win.
It's hard to see the bad stuff that doesn't exist (we avoided it). So I'm noting the bad stuff I see right now, to look for its absence in the future.

> I don't even want to imagine the nightmare we have avoided.

I need to imagine it. I need to take inventory of all benefits and costs. But imagination is limited, so one day I'd like to try out both ways. The feature flag mess we saw in other places could be avoided with reasonable care. Remember Jira API? The concept of API wasn't at fault, it was the implementation.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
dagguh: > How?

1. Deprecate
2. Release
3. Safe-rollout // not needed, because we can't `if` around a new class or around a `@Deprecated`, compile-time safety must be enough
4. Remove
5. Release // not needed, removing a class brings no value to the consumer, why would they bump for that?
6. Safe-rollout // not needed, because there's no release and also point 3. remains

> No. but It's pre 1.0.0. It can be removed anytime.

Correct, that's SemVer rule 4 👍

afaruga-atlassian:
> Deprecation is not required, but we should keep the changelog.

afaruga-atlassian:
> Deprecation is not required, but we should keep the changelog.

afaruga-atlassian:
> I wonder whether this shouldn't be blocked till #57 will be merged?
> So quick develoop would stick to concumer-devloop-cycle
> 
> If not then changes would accumulate and if 0.1.30 will be broken and 0.1.31 would be merged before those problems revealed, then 0.1.31 will have those bugs, and revert will have to apply for the booth of them. This introduces some juggling between versions, what will be overhead.
> 
> Do I get it correctly ?
> If yes it would be great to add some automation and block releasing new versions till the previous ones are not confirmed as compatible.


Thanks  @wyrzyk for discussion.

Let me share the short diagram I've made, I believe it might be useful for others.

So it's normal and expected that if bugs will be introduced then we would have to sorted them out and rolled out 2 changes at the same time, it will be good motivator to rollout smoothly and doesn't accumulated more changes at the same time.

An important constraint is:
* releasing version
* marking new version (blocking changelog) 
* cleanup in the product might be done by next dev who will upgrade db-replica just by changing the key of FF
Screenshot 2021-03-11 at 15 23 33

afaruga-atlassian (approved):

dagguh (approved):

PR #61, opened 2021-03-15
Mark release `1.0.0`

wyrzyk:
By approving the PR, you're approving the `1.0.0` API. I will merge it after the release.

- [x] Release `db-replica` 1.0.0 first
- [ ] Make sure Aurora LSN can be implemented with the current version of `ReplicaConsistency`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,10 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-0.1.31...master
+[Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.0.0...master
+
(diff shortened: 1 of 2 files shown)


dagguh:
> Make sure Aurora LSN can be implemented with the current version of `ReplicaConsistency`

It's a nice-to-have, but not a must. We can release 2.0.0.

dagguh (approved):
I'm ok with the current API, whatever flaws it might have. It's because [it's already on prod for a few releases already, so it should have been 1.0.0 for quite a while now](https://semver.org/#how-do-i-know-when-to-release-100).

wyrzyk:
> I'm ok with the current API, whatever flaws it might have. It's because it's already on prod for a few releases already, so it should have been 1.0.0 for quite a while now.

My only concern is that the only `ReplicaConsistency` implementation we were using is clock based and we don't consider it prod ready. But you're right we can release `2.0.0`.

wyrzyk:
https://packages.atlassian.com/maven-external/com/atlassian/db/replica/db-replica/1.0.0/

PR #62, opened 2021-03-17
new replica consistency mode: LSN (log sequence number)

tciesla-atlassian:


src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * It remembers last write LSN (log sequence number) and compares it with LSN for subsequent reads to detect inconsistencies.
+ * If it cannot remember the LSN of last write, pessimistically assumes it's going to be inconsistent.
+ *
+ * @see <a href="https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-global-database-monitoring.html">Monitoring Aurora PostgreSQL-based Aurora global databases</a>
+ * <p>
+ * It's an PostgreSQL-based Aurora database engine specific implementation.
+ */
+public class AuroraLsnReplicaConsistency implements ReplicaConsistency {
```
wyrzyk, line 22: It's API. Please declare it as final. https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#2314-consider-declaring-the-class-as-final

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private AuroraLsnReplicaConsistency(Cache<Long> lastWrite) {
+        this.lastWrite = lastWrite;
+    }
+
+    @Override
+    public void write(Connection main) {
+        try {
+            lastWrite.put(queryMainDbLsn(main));
+        } catch (Exception e) {
```
wyrzyk, line 53: I've checked the app-side, and we log the exception on the cache level before we re-throw it to the Replica consistency, so we are not losing the exception. I just wonder why to catch and handle it here. We don't do it in PessimisticPropagationConsistency. 🤔

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        return lastRefresh >= maybeLastWrite.get();
+    }
+
+    /**
+     * @return LSN (log sequence number) for the main database (the highest LSN)
+     */
+    private long queryMainDbLsn(Connection connection) throws Exception {
+        return queryLsn(connection, "SELECT MAX(durable_lsn) AS lsn FROM aurora_global_db_instance_status();");
```
wyrzyk, line 79: We are checking it for the current connection/replica. What if max(LSN) for one replica will be up to date and max LSN for the current one will not. Do we have a test to cover it?

wyrzyk: Ok, I see it's for the main database 🤣 .

wyrzyk: AFAIR in Postgres, it was more lightweight as we could select LSN instead of aggregate it.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * @return LSN (log sequence number) for the most outdated replica database (the lowest LSN)
```
wyrzyk, line 83: Can't we obtain the LSN of the current connection? One lagging replica could make all the others not serve the load.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+        return queryLsn(connection, "SELECT MIN(durable_lsn) AS lsn FROM aurora_global_db_instance_status();");
+    }
+
+    private long queryLsn(Connection connection, String rawSqlQuery) throws SQLException {
+        try (
+            PreparedStatement query = connection.prepareStatement(rawSqlQuery);
+            ResultSet results = query.executeQuery()
+        ) {
+            results.next();
+            return results.getLong("lsn");
```
wyrzyk, line 95: I remember we had to parse it in Postgres. Aurora seems to be more straightforward.

tciesla-atlassian: Exactly, in postgres LSN has text format like ''16/B374D848". Aurora returns just long type.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
afaruga-atlassian, line 22: We came from product assumption that Aurora is used with Postgress, if other product would came with Aurora-MySql than it would probably add `AuroraMySqlLsnReplicaConsistency`

we can either continue with default: Aurora -> Postgres-Aurora to keep it concise or we can enforce it : `AuroraPostgresLsnReplicaConsistency`

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: Responses from function `aurora_global_db_instance_status` contain field `server_id ` so i guess it's possible to match current connection with appropriate LSN. For now I can't see how responses look like because we don't have any aurora instance on staging and responses on prod instances are hashed.

I agree that one lagging replica shouldn't block others, but for know we only have tenants with only one replica. This issue will appear when we will scale up replicas. IMO we can descope it to separate task. WDYT?

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: True, I will rename it to `AuroraPostgresLsnReplicaConsistency `.

wyrzyk: I think the API we use will work the same way for AuroraPostgres and AuroraMySql.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
wyrzyk: Can you create a GitHub issue?

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @see <a href="https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-global-database-monitoring.html">Monitoring Aurora PostgreSQL-based Aurora global databases</a>
+ * <p>
+ * It's an PostgreSQL-based Aurora database engine specific implementation.
+ */
+public class AuroraLsnReplicaConsistency implements ReplicaConsistency {
+
+    private final Cache<Long> lastWrite;
+
+    public static class Builder {
```
wyrzyk, line 26: It's API and also could be final 😉 .

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: Right, using "aurora" function is more convenient. In the spare time we can also add version with "postgres" lsn and compare them. "Postgres" solution has this downside that is highly dependent on postgres version - functions names are different for different postgres versions. I guess "aurora" api will be more stable.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
afaruga-atlassian, line 26: Most often Builders are just an additional, verbose code, put at the bottom of the class.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
wyrzyk: We removed the Postgres one from the API to [avoid dependency on Postgres JDBC](https://github.com/atlassian-labs/db-replica/issues/9). We were using it to parse LSN.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: It's because `clock.instant()` is much less error-prone than `queryMainDbLsn(main)` :)

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: I added Builder in this same place as in class `PessimisticPropagationConsistency` to keep convention.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
afaruga-atlassian: So how to avoid dependencies when implementing LSN ?
LSN is dependent on database

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
wyrzyk: Yes. But here, we wrap `put` as well.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
wyrzyk: 1. Aurora provided LSN as long already, so no need to depend on a parser.
2. We can write our own Postgres `LSN to long` parser.
3. We can create another project `postgres-db-replica` with the extra dependency on Postgres JDBC. It will not block `db-replica` re-use for projects that don't need Postgres JDBC dependency.
4. We can implement PostgresLSN consistency in the product (ReplicaConsistency is an SPI).

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: In the case of put failure we should also invalidate cache. Why it shouldn't be part of consistency implementation?

wyrzyk: I don't say it shouldn't be here. I just noticed we didn't do it in the previous implementation.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: For now it looks that "aurora" way is only supported for PostgreSQL-based aurora db.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
wyrzyk: IMO put failure may mean it timeouts, and we don't know if it propagated. Then we can try to reset it. It may fail or not. I'm not sure what value we get from it.

Anyway, we should know the value we can document if with a test.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: Ok, if i understood correctly "aurora" solution is better because we avoid adding dependency on Postgres JDBC.

src/main/java/com/atlassian/db/replica/api/AuroraLsnReplicaConsistency.java
tciesla-atlassian: Done, #63

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    private long queryMainDbLsn(Connection connection) throws Exception {
+        return queryLsn(connection, "SELECT MAX(durable_lsn) AS lsn FROM aurora_global_db_instance_status();");
+    }
+
+    /**
+     * @return LSN (log sequence number) for the most outdated replica database (the lowest LSN)
+     */
+    private long queryReplicaDbLsn(Connection connection) throws Exception {
+        return queryLsn(connection, "SELECT MIN(durable_lsn) AS lsn FROM aurora_global_db_instance_status();");
```
dagguh, line 86: The `durable_lsn`:
> The LSN made durable in storage. 

If replica dies and gets restarted, it will have this LSN. But `DualConnection` doesn't care about durability, but visibility. So we should query `highest_lsn_rcvd`. Right?

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private long queryLsn(Connection connection, String rawSqlQuery) throws SQLException {
+        try (
+            PreparedStatement query = connection.prepareStatement(rawSqlQuery);
+            ResultSet results = query.executeQuery()
```
dagguh, line 92: They're auto-closed, nice!

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        long lastRefresh;
+        try {
+            lastRefresh = queryReplicaDbLsn(replica.get());
+        } catch (Exception e) {
+            return false;
+        }
+
+        return lastRefresh >= maybeLastWrite.get();
```
dagguh, line 72: `>=` is crucial, because `==` would reduce the replica ratio. Great to see it covered by the `shouldBeConsistentWhenReplicaLsnGreaterThanLastWriteLsn` test 👍

src/test/java/com/atlassian/db/replica/internal/AuroraPostgresLsnReplicaConsistencyTest.java
```diff
@@ -0,0 +1,117 @@
...
+        mockLsnFetching(replica, LAST_WRITE_LSN + 1);
+        consistency.write(main);
+
+        boolean isConsistent = consistency.isConsistent(new ConnectionSupplier(replica));
+
+        assertThat(isConsistent).isTrue();
+    }
+
+    @Test
+    public void shouldBeConsistentWhenReplicaLsnEqualToLastWriteLsn() throws Exception {
```
dagguh, line 94: Tests can use more business-level language instead of the technical implementation details:
```suggestion
    public void shouldBeConsistentWhenReplicaCaughtUp() throws Exception {
```

Accompanying test cases:
`shouldBeConsistentWhenReplicaIsAhead`
`shouldBeInconsistentWhenReplicaIsBehind`

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * It remembers last write LSN (log sequence number) and compares it with LSN for subsequent reads to detect inconsistencies.
+ * If it cannot remember the LSN of last write, pessimistically assumes it's going to be inconsistent.
+ *
+ * @see <a href="https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-global-database-monitoring.html">Monitoring Aurora PostgreSQL-based Aurora global databases</a>
+ * <p>
+ * It's an PostgreSQL-based Aurora database engine specific implementation.
+ */
+public final class AuroraPostgresLsnReplicaConsistency implements ReplicaConsistency {
```
dagguh, line 22: New API is of interest to consumers, so the changelog needs to mention it

src/main/java/com/atlassian/db/replica/api/PessimisticPropagationConsistency.java
```diff
@@ -14,13 +14,13 @@
  * Assumes that writes propagate from main to replicas in at most a given amount of time.
  * If it cannot remember the time of last write, pessimistically assumes it's going to be inconsistent.
  */
-public class PessimisticPropagationConsistency implements ReplicaConsistency {
+public final class PessimisticPropagationConsistency implements ReplicaConsistency {
```
dagguh, line 17: Technically this breaks source compatibility. Subclasses will fail compilation now. So we should mark it in the changelog and bump a major version.

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/api/PessimisticPropagationConsistency.java
wyrzyk: IMO we should leave it as is. There's no need to change it now.

dagguh: Kotlin made me lazy, forgot that I have to opt-in to maintainability keywords in Java 😉

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
tciesla-atlassian: I didn't found good explanation of differences between fields, so this is my understanding:
- durable_lsn (it's persisted and served)
- highest_lsn_rcvd (it's received, not persisted and not served yet?)
- oldest_read_view_lsn (lsn during last read operation)
I have doubts whether `highest_lsn_rcvd` means that those changes are served to clients. @dagguh WDYT?

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private AuroraPostgresLsnReplicaConsistency(Cache<Long> lastWrite) {
+        this.lastWrite = lastWrite;
+    }
+
+    @Override
+    public void write(Connection main) {
+        try {
+            lastWrite.put(queryMainDbLsn(main));
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
tciesla-atlassian: What do you suggest here? IMO we should add logger here but it's not yet added to library.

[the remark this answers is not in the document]
dagguh: Adding a logger (even as generic as slf4j) adds friction for the consumers, they'd have to:
* bridge their logging framework to ours (it's easy but not zero cost)
* synchronize on logging dependency version
* force English exception messages

Instead, we can rethrow and let the consumer choose their error handling (logging, retrying, etc.). The `reset` can fail too, so they'll have to do it anyway.

External retries would work on `write(Connection main)` level, rather than just the `queryMainDbLsn` part. If this granularity is insufficient, we can add `.put(Supplier<T> value)` method to the cache. But I'd wait with that until it's actually needed.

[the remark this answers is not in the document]
wyrzyk: > Adding a logger
Postgres JDBC uses `java.util.logging` to avoid adding dependencies.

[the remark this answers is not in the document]
tciesla-atlassian: I'm not sure that we should rethrow exception and force api client to handle it. We know that in case of any error we should invalidate cache to make further reads inconsistent. What is missing is only part about "visibility" so maybe it's enough to log it using `java.util.logging`?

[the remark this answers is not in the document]
dagguh: JUL could work, nice find. But the `reset` itself is not so clear-cut.

The `reset` call can also fail, so it doesn't guarantee inconsistency passing anyway. There are also additional costs to this.

If cache is timing out, then `put` was already slow. Then we call `reset`, which will slow it down even more and probably time out too.

Also this is a check-then-act race condition, which can erase a legitimate new `lastWrite` from another request.

We shouldn't force this error handling. Let's ship the LSN part itself and leave error handling for another place. The consumer can choose to reset if they want in their own `ReplicaConsistency` layer (or in `Cache` if it gets `put` with `Supplier`).

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
tciesla-atlassian: `highest_lsn_rcvd ` in case of main db is always null:
Screenshot 2021-03-17 at 15 39 10

tciesla-atlassian: so if "durable" seems to be lower bound for "visible" then it's the safest option here.

wyrzyk: Make sense. Maybe we can use it instead of aggregation to distinguish between main and replicas.

tciesla-atlassian: @wyrzyk Will it give us much performance gain (`WHERE hightest_lsn_rcsv` != NULL`) than aggregation with `MIN/MAX`?

dagguh: Interesting. Is there an alternative call which shows LSN? Maybe the Postgres function?

It's also interesting to see that replica `highest_lsn_rcvd` is bigger than its `durable_lsn`, which seems to confirm my assumption. In this case we can see that replica received the latest durable LSN from main.
Maybe main only has durable LSN because it doesn't "receive" LSN, it is sending it.

So here's my proposal:
* read `durable_lsn` from main
* read `highest_lsn_rcvd` from replica

wyrzyk: > Will it give us much performance gain (WHERE hightest_lsn_rcsv != NULL) than aggregation with MIN/MAX?

I have no idea.

tciesla-atlassian: @dagguh Whole point of LSN usage is to avoid consistency errors. We are not sure whether `highest_lsn_rcvd ` will serve changes between `durable_lsn ` and `highest_lsn_rcvd` so it can introduce consistency issue. IMO we should use  `durable_lsn ` for both.

dagguh: Very good point. That's the safe way 👍

dagguh: So the current code is correct

dagguh (dismissed):

wyrzyk (approved):

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
dwatl, line 22: AuroraPostgres -> Aurora?

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public boolean isConsistent(Supplier<Connection> replica) {
+        Optional<Long> maybeLastWrite = lastWrite.get();
+        if (!maybeLastWrite.isPresent()) {
+            return false;
+        }
+
+        long lastRefresh;
```
dwatl, line 65: We can inline it

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+    public void write(Connection main) {
+        try {
+            lastWrite.put(queryMainDbLsn(main));
+        } catch (Exception e) {
+            throw new RuntimeException("failure during LSN fetching for main database", e);
+        }
+    }
+
+    @Override
+    public boolean isConsistent(Supplier<Connection> replica) {
```
dwatl, line 59: Let's try to make it more concise/idiomatic?

dwatl (approved):

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
dwatl: We could have this, but exceptions are not helping. Not sure if inlining exception handling to `queryReplicaDbLsn` makes it more readable.

```suggestion
    public boolean isConsistent(Supplier<Connection> replica) {
        return lastWrite.get()
            .map(lw -> queryReplicaDbLsn(replica.get()) >= lw)
            .orElseGet(false);
    }
```

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -0,0 +1,99 @@
...
+    private AuroraPostgresLsnReplicaConsistency(Cache<Long> lastWrite) {
+        this.lastWrite = lastWrite;
+    }
+
+    @Override
+    public void write(Connection main) {
+        try {
+            lastWrite.put(queryMainDbLsn(main));
+        } catch (Exception e) {
+            throw new RuntimeException("failure during LSN fetching for main database", e);
```
dwatl, line 54: Do we really want to throw here? We shouldn't expect writes will always work, they might timeout etc.

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
dagguh: There's a [resolved convo about this](https://github.com/atlassian-labs/db-replica/pull/62#discussion_r595922980).

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
dagguh: What else can we do?

Relevant [convo1](https://github.com/atlassian-labs/db-replica/pull/62#discussion_r595895428) and [convo2](https://github.com/atlassian-labs/db-replica/pull/62#discussion_r595946770). You can unresolve and comment there if you want to reuse the context.

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
dagguh: To unblock functional style, we'd have to get rid of the checked exceptions from `queryReplicaDbLsn`. We could convert checked -> unchecked there.

dwatl: Maybe this 🤔 
```suggestion
        public boolean isConsistent(Supplier<Connection> replica) {
        try {
            long lsn = queryReplicaDbLsn(replica.get());

            return lastWrite.get()
                .map(lw -> lsn >= lw)
                .orElse(false);
        } catch (SQLException e) {
            return false;
        }
    }
```

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
tciesla-atlassian: We can't without moving line `lastRefresh >= maybeLastWrite.get()` to try-catch block. IMO current version is better because at first glance it show that there are 3 sections with pre-checks and conditions.

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
tciesla-atlassian: Look that we don't need to fetch replica LSN when there is not LSN for lastWrite, so this version has worse performance...

tciesla-atlassian: Ok, it looks good:
`try {
            return lastWrite.get()
                .map(lastWriteLsn -> queryReplicaDbLsn(replica.get()) >= lastWriteLsn)
                .orElse(false);
        } catch (Exception e) {
            return false;
        }`
        
WDYT?

dagguh (approved):

dwatl (approved):
LGTM :)

PR #64, opened 2021-03-18
Clarify the release procedure

dagguh:


docs/release/releasing.md
```diff
@@ -1,19 +1,17 @@
 # Releasing
+1. Ensure that the next version maintains SemVer:
+    * Check if `Unreleased` changes need a MINOR or MAJOR bump.
+    * If yes, then mark the next version as per [`gradle-release` plugin] docs.
```
wyrzyk, line 4: IMO we can automate it in Github Actions.

docs/release/releasing.md
```diff
@@ -1,19 +1,17 @@
...
+    * Otherwise, it will default to a PATCH bump.
```
wyrzyk, line 5: We could stop using default and make it required in GH actions instead of the `yes/no` param.

wyrzyk (approved):

docs/release/releasing.md
dagguh: Ideally, each pull request should be able to bump it, but tags cannot be parts of PRs unfortunately.
So GHA could have an input with `version: 1.1.0` or with `bump:minor`.

wyrzyk: IMO it's just git tag `version-alpha` && git push origin version-alpha

dagguh: Or we could say bye bye to axion plugin and hold the versions in the code rather than tags.

dagguh: > IMO it's just git tag version-alpha && git push origin version-alpha

Locally I also use `git` directly, circumventing the `./gradlew markNextVersion`. Documenting the command would be quite verbose, because the `$version` is a variable and the `$remote` is `origin` for non-forkers and `upstream` for forkers (by convention).

dagguh: Axion has this `./gradlew markNextVersion -Prelease.incrementer=incrementMinor` ✨ 
I'll go for the `bump` input in GHA

dagguh: Done, although the video is out of date now 😁
The GUI flow is the same, but the example values are different

.github/workflows/ci.yml
```diff
@@ -31,10 +31,11 @@ jobs:
         name: test-reports
         path: build/reports/tests
     - name: Release
-      if: github.event.inputs.release == 'yes'
+      if: github.event.inputs.release != 'None'
       env:
         atlassian_private_username: ${{ secrets.ARTIFACTORY_USERNAME }}
         atlassian_private_password: ${{ secrets.ARTIFACTORY_API_KEY }}
       run: |
+        ./gradlew markNextVersion -Prelease.incrementer=increment$ {{ github.event.inputs.release }} -Prelease.localOnly
```
wyrzyk, line 39: Does it work?

dagguh: It worked locally for me. We'll know the full flow if we actually release 1.1.0.
@tciesla-atlassian wanna try or should I test my own contraptions? 😁

tciesla-atlassian (approved):

docs/release/releasing.md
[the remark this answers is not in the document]
dagguh: I deliberately didn't want to copy examples from another repo. I'd have to synchronize it if it changes. But in practice it didn't change for a long long time, so maybe locality would be worth a bit of denormalization.

If the GHA incrementer doesn't work, I'll add the example.

afaruga-atlassian:
Folks you're too fast! I had a meeting in meantime and didn't have a chance to review & approve this one.
**Good stuff**

.github/workflows/ci.yml
dagguh: Tomek bravely tested it and we got [a green 1.1.0 release](https://github.com/atlassian-labs/db-replica/runs/2140369424) 😎

PR #66, opened 2021-03-22
Impl missing DualConnection methods

dwatl:


src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -375,25 +375,45 @@ public PreparedStatement prepareStatement(String sql, String[] columnNames) thro
     @Override
     public Clob createClob() throws SQLException {
         checkClosed();
-        throw new ReadReplicaUnsupportedOperationException();
+        if(compatibleWithPreviousVersion)
+            throw new ReadReplicaUnsupportedOperationException();
+        else
+            return connectionProvider
+                .getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL))
```
wyrzyk, line 382: Why write-connection?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -375,25 +375,45 @@ public PreparedStatement prepareStatement(String sql, String[] columnNames) thro
     @Override
     public Clob createClob() throws SQLException {
         checkClosed();
-        throw new ReadReplicaUnsupportedOperationException();
+        if(compatibleWithPreviousVersion)
+            throw new ReadReplicaUnsupportedOperationException();
+        else
+            return connectionProvider
```
dagguh, line 381:
```suggestion
        if (compatibleWithPreviousVersion) {
            throw new ReadReplicaUnsupportedOperationException();
        } else {
            return connectionProvider
```

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -375,25 +375,45 @@ public PreparedStatement prepareStatement(String sql, String[] columnNames) thro
...
 
     @Override
     public Blob createBlob() throws SQLException {
         checkClosed();
-        throw new ReadReplicaUnsupportedOperationException();
+        if(compatibleWithPreviousVersion)
+            throw new ReadReplicaUnsupportedOperationException();
+        else
+            return connectionProvider
+                .getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL))
```
dagguh, line 393: Let's add the fix to the changelog

dagguh (changes requested):

wyrzyk:
No test no fix. :trollface:

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -482,7 +517,12 @@ public void setNetworkTimeout(Executor executor, int milliseconds) throws SQLExc
     @Override
     public int getNetworkTimeout() throws SQLException {
         checkClosed();
-        throw new ReadReplicaUnsupportedOperationException();
+        if(compatibleWithPreviousVersion)
+            throw new ReadReplicaUnsupportedOperationException();
+        else
+            return connectionProvider
+                .getReadConnection(new RouteDecisionBuilder(Reason.RO_API_CALL))
```
afaruga-atlassian, line 524: You didn't add tests, So I guess that you choose to add `RO_API_CALL` `RW_API_CALL` based on some product usage. Can you add tests ?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -375,25 +375,45 @@ public PreparedStatement prepareStatement(String sql, String[] columnNames) thro
     @Override
     public Clob createClob() throws SQLException {
         checkClosed();
-        throw new ReadReplicaUnsupportedOperationException();
+        if(compatibleWithPreviousVersion)
+            throw new ReadReplicaUnsupportedOperationException();
```
afaruga-atlassian, line 379: How do you want to roll out and test those changes? Do you have product that will use them ? AFAIK we don't have any way to verify the rollout, by verify I meant use the implementation on production.

wyrzyk: I guess any implementation will work better than `throw new ReadReplicaUnsupportedOperationException();`. This implementation delegates to a real connection so should be safe. The only question can it use a `read` connection or do we have to use a `write` connection?
I think the main goal is to slowly implement the methods we don't support and remove `ReadReplicaUnsupportedOperationException`.

dagguh: Using `getWriteConnection` will work for sure.
Other methods use `getReadConnection`, which would throw `SQLException` if mis-routed. Tracking `ReadReplicaUnsupportedOperationException` could actually be easier than yet another `SQLException`, but we have to try using replica at some point.

wyrzyk: We can use `getWriteConnection` as a safe option and think about switching when proofed it hurts efficiency.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: IMO it's noisier, but I will follow the convention

wyrzyk: It's `pythonier`.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
afaruga-atlassian: > I guess any implementation will work better than

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: What would you like to test here? 🤔

wyrzyk: We can test if the delegate has been called, as the implementation seems to be just a delegate. Otherwise, the method will not be test-covered. Not sure if it's worth it.

dagguh: We also didn't integration-test every single JDBC method, because we don't have a JDBC/SQL workload covering all these APIs.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: Maybe it's the same?
image

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: Tests based on implementation are bad practice or at best very low value. It makes implementation harder to change, even when logic stays the same.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: I added commit message to explain reasoning behind this

dagguh: It's not pushed yet

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: How do other open-source libraries do it? They simply wait for feedback from consumers and release bugfixes.
Any reason why this strategy won't work for us?

afaruga-atlassian: Agree.

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ### Added
 - `AuroraPostgresLsnReplicaConsistency`
 
+### Fixed
+- Implemented missing methods in `DualConnection`
```
afaruga-atlassian, line 33: NIT: I wonder why you adding it under `1.1.0` version ? IMHO you should add it under the new version, in _mark commit ( to update changelog and readme) _

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -482,7 +524,13 @@ public void setNetworkTimeout(Executor executor, int milliseconds) throws SQLExc
     @Override
     public int getNetworkTimeout() throws SQLException {
         checkClosed();
-        throw new ReadReplicaUnsupportedOperationException();
+        if(compatibleWithPreviousVersion) {
+            throw new ReadReplicaUnsupportedOperationException();
+        } else {
+            return connectionProvider
+                .getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL))
```
afaruga-atlassian, line 531: I'm still missing any sanity tests for all of those methods.
* It'll prevent regression this implementation,
* will be the starting point for any other changes related to those methods
* sustain current level of code coverage

afaruga-atlassian (changes requested):

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dagguh: Postgres JDBC driver has a [huge suite of tests](https://github.com/pgjdbc/pgjdbc/blob/v42.3.0-rc5/pgjdbc/src/test/java/org/postgresql/test/jdbc2/Jdbc2TestSuite.java).

CHANGELOG.md
dagguh: Good catch, this should be under `Unreleased`

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dagguh: > How do other open-source libraries do it? They simply wait for feedback from consumers and release bugfixes.

Which lib does this? Every repo I saw asked contributors to include tests.

dwatl: What cases would you like to see being tested?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dwatl: What cases/behavior would you like to see being tested?

src/main/java/com/atlassian/db/replica/api/DualConnection.java
dagguh: Smoke test by calling these methods in `DualConnectionIT`. It can be a single test case going through all the methods.
The harness is already set up, it should be quite cheap to write.

wyrzyk:
@dwatl I added an integration test, cherry-pick your commit and implemented the rest of the DualConnection methods (https://github.com/atlassian-labs/db-replica/pull/72).

dwatl:
Closing in favour of https://github.com/atlassian-labs/db-replica/pull/72

PR #67, opened 2021-03-25
Add mutation tests

dwatl:
Output example
> Generated 1468 mutations Killed 516 (35%)
> Mutations with no coverage 768. Test strength 74%
> Ran 3296 tests (2.25 tests per mutation

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,16 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath("info.solidsoft.gradle.pitest:gradle-pitest-plugin:1.4.7")
+    }
+}
+
 plugins {
     `java-library`
     id("com.atlassian.performance.tools.gradle-release").version("0.7.3")
+    id("info.solidsoft.pitest").version("1.6.0")
 }
 
 tasks.wrapper {
@@ -63,6 +73,12 @@ tasks.test {
     }
 }
 
+val pitest = tasks.withType<info.solidsoft.gradle.pitest.PitestTask> {
+    threads.set(12)
+    outputFormats.set(setOf("XML", "HTML"))
+    avoidCallsTo.set(setOf("kotlin.jvm.internal", "kotlinx.coroutines"))
+}
+
 val testIntegration = task<Test>("testIntegration") {
     filter {
         include("**/*IT.class")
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -63,6 +73,12 @@ tasks.test {
     }
 }
 
+tasks.withType<info.solidsoft.gradle.pitest.PitestTask> {
```
dagguh, line 76: Let's make `check` task depend on this one

build.gradle.kts
```diff
@@ -63,6 +73,12 @@ tasks.test {
...
+    threads.set(12)
+    outputFormats.set(setOf("XML", "HTML"))
```
dagguh, line 78: Let's add these reports as GitHub Actions artifacts.

dagguh (changes requested):

build.gradle.kts
dagguh: Ok, `check` runs `pitest`:
```
» ./gradlew check --dry-run
Starting a Gradle Daemon (subsequent builds will be faster)
:compileJava SKIPPED
:processResources SKIPPED
:classes SKIPPED
:compileTestJava SKIPPED
:processTestResources SKIPPED
:testClasses SKIPPED
:test SKIPPED
:pitest SKIPPED
:testIntegration SKIPPED
:verifyLicensing SKIPPED
:check SKIPPED
```

dagguh (approved):

build.gradle.kts
```diff
@@ -63,6 +73,12 @@ tasks.test {
     }
 }
 
+val pitest = tasks.withType<info.solidsoft.gradle.pitest.PitestTask> {
```
dagguh, line 76: Tests failed:
```
10:24:00 AM PIT >> INFO : MINION : java.lang.AssertionError: [thruput per ms] 
10:24:00 AM PIT >> INFO : MINION : Expecting:
10:24:00 AM PIT >> INFO : MINION :  <1978.9049f>
10:24:00 AM PIT >> INFO : MINION : to be greater than:
10:24:00 AM PIT >> INFO : MINION :  <2500.0f> 
```
It seems the mutations are slowing down the benchmark. I guess we can drop the threshold?

dagguh: This plugin is spamming the stdout. A `./gradlew build` should be basically empty unless there's an actionable for the dev, like a red test or a warning.

- [ ] reduce/remove stdout spam

dagguh (changes requested):

build.gradle.kts
dwatl: We could exclude test from mutation. It's not checking the logic, but performance. Mutations are more useful for checking uncovered logical cases.

PR #68, opened 2021-03-30
Extend Aurora lsn with cache

dwatl:


dwatl:
Gonna update changelog real quick

CHANGELOG.md
```diff
@@ -27,7 +27,8 @@ Dropping a requirement of a major version of a dependency is a new contract.
 [1.1.0]: https://github.com/atlassian-labs/db-replica/compare/release-1.0.0...release-1.1.0
 
 ### Added
-- `AuroraPostgresLsnReplicaConsistency`
+- `KeyBasedThrottledCache`
+- `SuppliedCache<T>`
```
tciesla-atlassian, line 31: I should be added in `unreleased` section... not in `1.1.0 - 2021-03-18` section

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -33,16 +38,28 @@ public Builder cacheLastWrite(Cache<Long> lastWrite) {
             return this;
         }
 
+        /**
+         * @param cache
+         */
```
tciesla-atlassian, line 43: please add more robust comment or drop it completely :)

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
+
+/**
+ * Every read can potentially refresh the cache, however it can be loaded only by one thread at the time.
+ * Other threads instead of waiting are getting last known value.
+ */
+public class KeyBasedThrottledCache implements SuppliedCache<Long> {
```
tciesla-atlassian, line 14: maybe it should be named in some simpler way like: KeyBasedNoConcurrentRefreshCache? WDYT?

src/main/java/com/atlassian/db/replica/api/LsnReplicaConsistency.java
```diff
@@ -0,0 +1,86 @@
...
+import com.atlassian.db.replica.spi.Cache;
+import com.atlassian.db.replica.spi.ReplicaConsistency;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.function.Supplier;
+
+public final class LsnReplicaConsistency implements ReplicaConsistency {
```
tciesla-atlassian, line 13: Why do we need this class? We have `AuroraPostgresLsnReplicaConsistency ` already...

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
```diff
@@ -0,0 +1,110 @@
...
+        cache.get(() -> 3L);
+
+        Long lastValue = 4L;
+        Optional<Long> actual = cache.get(() -> lastValue);
+
+        assertThat(actual).hasValue(lastValue);
+    }
+
+    @Test
+    public void noConcurrentInvalidation() throws InterruptedException {
```
tciesla-atlassian, line 44: I don't caught the idea of this unit test. What is the assertion here? That there is no exception and empty optional is returned?

tciesla-atlassian (commented):
Looks good :)

src/main/java/com/atlassian/db/replica/api/AuroraPostgresLsnReplicaConsistency.java
```diff
@@ -33,16 +38,28 @@ public Builder cacheLastWrite(Cache<Long> lastWrite) {
...
+        public Builder replicaLsnCache(SuppliedCache<Long> cache) {
```
dagguh, line 44: It's a way of injecting a new feature via new API, without affecting old API. And the default is equivalent to old behavior. Very well done.

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
```diff
@@ -0,0 +1,43 @@
...
+    private static final ConcurrentHashMap<String, Long> lastValues = new ConcurrentHashMap<>();
```
dagguh, line 15: The `static` makes testing and GC harder.

If the keying aspect is decoupled or removed, the locking aspect will be easier to test in integration tests.

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ConcurrentModificationException;
+import java.util.Optional;
+import java.util.Random;
+import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class KeyBasedThrottledCacheTest {
```
dagguh, line 14: Good test cases 👍

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
```diff
@@ -0,0 +1,110 @@
...
+        String key = randomKey();
+        KeyBasedThrottledCache cache = new KeyBasedThrottledCache(key);
+        KeyBasedThrottledCache cache2 = new KeyBasedThrottledCache(key);
+
+        cache.get(() -> 1L);
+
+        asyncCacheUpdateWithSlowSupplier(cache, 2L);
+        asyncCacheUpdateWithSlowSupplier(cache2, 3L);
+
+        Optional<Long> actual = cache2.get(this::anyValue);
```
dagguh, line 64: It's magical how the `cache` and `cache2` interact due to the hidden global state.

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
dagguh: The loaded name is caused by having two responsibilities: keying per String + locking.

I infer the `String key` will represent a replica JDBC url or a tenant id. We already have a cross-replica and cross-tenant in-memory state: the `AbstractConnectionPoolProvider`. It can hold the `ConcurrentHashMap<String, SuppliedCache>`.
And the `SuppliedCache` itself would become just a `ReentrantLock` + a value field.

Alternatively, the `ConcurrentHashMap` aspect could be decoupled from the `if (lock.tryLock()` aspect as individual `SuppliedCache` impls.

dagguh (changes requested):

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
dagguh: Exactly. We can make it explicit via AssertJ: `Assertions.catchThrowable`

dwatl: It's pretty much proving that no concurrent invalidation is happening. If you scrap the locking it will fail.

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class KeyBasedThrottledCacheTest {
+    private final ExecutorService executor = Executors.newFixedThreadPool(10);
+
+    @Test
+    public void staleWhileInvalidating() throws InterruptedException {
```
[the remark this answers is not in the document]
dwatl: I only cache replica LSN. Staleness = potential inconsistency.

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
```diff
@@ -0,0 +1,43 @@
...
+     * @return last known value or null
+     */
+    public Optional<Long> get(Supplier<Long> supplier) {
+        maybeRefresh(supplier);
+        return Optional.ofNullable(lastValues.get(key));
+    }
+
+    private void maybeRefresh(Supplier<Long> supplier) {
+        ReentrantLock lock = locks.get(key);
+        if(lock.tryLock()) {
```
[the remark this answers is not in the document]
dwatl: Lower DB usage is the biggest gain. Skipping the supplier when it's heavy = a huge win (🤞 ).

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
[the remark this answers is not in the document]
dagguh: Aaah I see. So the worst the staleness can do is to route to main. Nice!
In short this is a tradeoff between "query LSN a lot" and "route more to replica". Can't wait to see the metrics 🤞

src/test/java/com/atlassian/db/replica/api/KeyBasedThrottledCacheTest.java
dwatl: `        cache.get(() -> 1L);
        asyncCacheUpdateWithSlowSupplier(cache, anyValue());

        cache2.get(this::anyValue);
        Optional<Long> actual = cache2.get(this::anyValue);

        assertThat(actual).hasValue(1L);`
        
        Is this easier to read?

dwatl: It's proving that when **cache1 blocks**, **cache2 is getting stale value from cache1**. 

_Detail: initial impl didn't need slow supplier for cache 2._

dagguh: The magic is an argument in the [keying discussion](https://github.com/atlassian-labs/db-replica/pull/68/files#r604088954). If keyed global state will remain, then the test is good as is. It accurately represents true behavior.

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
dagguh: Let me gather all the consequences of `static` state:
* [testing and GC](https://github.com/atlassian-labs/db-replica/pull/68/files#r604083839)
* [limited to `Long`](https://github.com/atlassian-labs/db-replica/pull/68/files#r604068028)
* [magical implicit interactions](https://github.com/atlassian-labs/db-replica/pull/68/files#r604086184)
* [two responsibilities](https://github.com/atlassian-labs/db-replica/pull/68/files#r604088954)

Each on its own might be a nit-pick. But together it seems worth trying to get rid of `static`. Or at least isolate it from the valuable throttling feature.

tciesla-atlassian: I also really like the idea of extracting key functionality and keeping only lock + value field here. Really nice hint Maciej!

wyrzyk:
Can we start from the internal implementation and then promote it to the library?
- `AuroraPostgresLsnReplicaConsistency` didn't work on production, and we are working on a new version internally. This one will probably be deprecated.
- I think it can expand to more than a cache. We already have locks here so it enables us to create 'pollingless' implementation of 'wait-for-consistency' strategy.

wyrzyk:
OTOH I see we are fixing [AuroraPostgresLsnReplicaConsistency](https://github.com/atlassian-labs/db-replica/pull/69) so the cache may stay here as well.

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -0,0 +1,36 @@
...
+
+
+/**
+ * Every read can potentially refresh the cache, however it can be loaded only by one thread at the time.
+ * Other threads instead of waiting are getting last known value.
+ */
+@ThreadSafe
+public class ThrottledCache<T> implements SuppliedCache<T> {
+    private final ReentrantLock lock = new ReentrantLock();
+    private final AtomicReference<T> value = new AtomicReference<>();
```
wyrzyk, line 19: Nit: We don't need AtomicReference here. The code doesn't use CAS.

dwatl:
I split the class, tests are still green. Gonna add/refactor tests and then do some impl refactoring.

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+/**
+ * Allows for thread-safe key-based access to ThrottledCache.
+ * It can be used to have only one instance per key (e.g. database url).
+ */
+@ThreadSafe
+public class KeyBasedThrottledCache implements SuppliedCache<Long> {
+    private static final ConcurrentHashMap<String, ThrottledCache<Long>> caches = new ConcurrentHashMap<>();
```
dagguh, line 16:
```suggestion
    private static final ConcurrentHashMap<String, SuppliedCache<T>> caches = new ConcurrentHashMap<>();
```
Now it can work with any `SuppliedCache`, so we could inject `Supplier<SuppliedCache<T>>`, freeing the consumer to use different strategies while preserving the keying. And also holding different values than `Long`.

dagguh (dismissed):

tciesla-atlassian (dismissed):

dwatl:
I'm considering dropping KeyBasedThrottledCache altogether. It's hard to test in this form and I'm not sure about the value.
WDYT?

dagguh:
> I'm considering dropping KeyBasedThrottledCache altogether. It's hard to test in this form and I'm not sure about the value.
WDYT?

Yes, exactly!

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
wyrzyk, line 16: Still static 😢 .

src/main/java/com/atlassian/db/replica/api/KeyBasedThrottledCache.java
```diff
@@ -0,0 +1,27 @@
...
+    private final String key;
+
+    public KeyBasedThrottledCache(String key) {
+        this.key = key;
+        caches.putIfAbsent(key, new ThrottledCache<>());
```
wyrzyk, line 21: A new `ThrottledCache` will be created for every `KeyBasedThrottledCache`. Can you change it to `computeIfAbsent`?

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.1.0...master
 
+### Added
+- `KeyBasedThrottledCache`
+- `SuppliedCache<T>`
```
wyrzyk, line 28: Why do we need it? Do we plan to use different suppliers every call? Can't we use Cache and inject the supplier via constructor?

wyrzyk: And if the supplier can change from call to call, how can we trust the previous supplier compatible with ours (when we read the stale data)?

wyrzyk (changes requested):
Please explain the need for `SuppliedCache`.

dwatl:
@wyrzyk Suppliers are not taking inputs as argument (connection can be different on every call).

CHANGELOG.md
dwatl: Suppliers are not taking inputs as a argument. Connection in our impl can be different on every call.

dwatl: There are 2 impls, one is for backward compatibility in a sense.

wyrzyk: > Suppliers are not taking inputs as a argument. Connection in our impl can be different on every call.

Yes, but You can have multiple implementations of the supplier. If there will be only one implementation then we can inject it via the constructor.

wyrzyk: Ok, so there's one implementation of the supplier. It's just a way to couple `AuroraPostgresLsnReplicaConsistency` and the cache, as the cache needs to call the supplier from `AuroraPostgresLsnReplicaConsistency`. `AuroraPostgresLsnReplicaConsistency` can't create the cache and inject supplier by the constructor as it's provided by a user. 🤔 


1.

dwatl: Did you have a chance to look at test cases? These should be displaying we can't inject supplier via constructor.

wyrzyk: > Did you have a chance to look at test cases?

Yes. I've seen it. It doesn't explain why we can't inject the values via a constructor. I think the test cases are misleading because every call uses a different supplier, so it shouldn't accept stale data generated by a different supplier.

```
 public void staleWhileInvalidating() throws InterruptedException {
        KeyBasedThrottledCache cache = new KeyBasedThrottledCache(randomKey());
        Long staleValue = 1L;
        cache.get(() -> staleValue);

        asyncCacheUpdateWithSlowSupplier(cache, 2L);

        Optional<Long> actual = cache.get(() -> 3L);

        assertThat(actual).hasValue(staleValue);
    }
```

 `Optional<Long> actual = cache.get(() -> 3L);` returns 1 for the `() -> 3L` supplier, but it shouldn't ever return a value different than 3 for this supplier. It's an logical error. The data is not stale, it's incorrect.

wyrzyk: The correct test should use one supplier that increases the value for each call. Then it would show the data is stale, and it would show the supplier can be injected via constructor.

dwatl: In practice, we'll have different connections and potentially different values on each supplier call.
We would need a function `(connection) -> long` as a supplier to inject is via constructor.

wyrzyk: I think we can decouple throttling and cache and use a Cache instead of introducing `SuppliedCache`. 🤔

dwatl: How would you implement put and reset from com.atlassian.db.replica.spi.Cache so it makes sense?

dwatl: If the supplier signature is `() -> long` and we inject it via a constructor. How to provide different connections? (every thread has a different connection, but the cache is shared across multiple threads)

dwatl: > It's an logical error.
Only if cache.get would be a pure function, it depends on the external state. Hence it's hard to reason about 🤷  .

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -0,0 +1,75 @@
...
+        cache.get(this::anyValue);
+
+        assertThat(cache.get(() -> 4L)).hasValue(4L);
+    }
+
+    @Test
+    public void noConcurrentInvalidation() throws InterruptedException {
+        ThrottledCache<Long> cache = new ThrottledCache<>();
+
+        asyncPutWithSlowSupplier(cache, 2L);
```
afaruga-atlassian, line 40: assertion For `2L` ?

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.1.0...master
 
+### Added
+- `KeyBasedThrottledCache`
```
tciesla-atlassian, line 27:
```suggestion
- `ThrottledCache`
```

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+/**
+ * Every read can potentially refresh the cache, however it can be loaded only by one thread at the time.
+ * Other threads instead of waiting are getting last known value.
+ */
+@ThreadSafe
+public class ThrottledCache<T> implements SuppliedCache<T> {
+    private final ReentrantLock lock = new ReentrantLock();
+    private T value = null;
```
tciesla-atlassian, line 18: Assigning null here is not necessary. If your intention was to emphasise it then it can stay ;)

tciesla-atlassian (dismissed):

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
tciesla-atlassian: I guess it doesn't matter, whole point here is to prove that no exception will be thrown.

dwatl: TBH it should be anyValue - not important for test case.

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
dwatl: Implicit defaults are easier to read

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.ConcurrentModificationException;
+import java.util.Random;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
+
+public class ThrottledCacheTest {
+    private final ExecutorService executor = Executors.newFixedThreadPool(10);
```
afaruga-atlassian, line 14: why 10 ?

afaruga-atlassian, line 14:
```suggestion
    private final ExecutorService executor = Executors.newSingleThreadExecutor()
```

afaruga-atlassian: I'm lacking test with  spawning 10 threads and testing concurrent blocking on `lock.tryLock()`

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        assertThat(cache.get(this::anyValue)).isEmpty();
+    }
+
+
+    private void asyncPutWithSlowSupplier(ThrottledCache<Long> cache, long value) throws InterruptedException {
+        executor.submit(() -> {
+            cache.get(() -> {
+                try {
+                    Thread.sleep(1000);
```
afaruga-atlassian, line 62: I've changed it to 1000  -> 100, and below  100-> 10 and all tests passed.

dwatl: Did you re-run 1k times? It may appear flaky on slower machines.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
dwatl: It's not really concurrent ;)

tciesla-atlassian (dismissed):

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
dwatl: Maybe unbounded?

CHANGELOG.md
wyrzyk: Yes, we work in context of `ReplicaConsistency#isConsistent(Supplier<Connection> replica)`. We need to pass the connection dynamically. I thought we might move throttling from the cache to ReplicaConsistency implementation (ThrottledConsistency), but there's no enough data in ReplicaConsistency scope. I think it needs to stay as-is.

wyrzyk (approved):

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
afaruga-atlassian: that's my point it can be simplified,

afaruga-atlassian: You could add a test with 10 requests, without timeout and CountDownLatch in between.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
afaruga-atlassian: My point is that you might use countdown latch to test it deterministically. thought its okish

afaruga-atlassian (approved):

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
dwatl: I changed the 100ms as it's affecting execution time. 1s is async time - we are not waiting for it.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
dwatl: Added something without arbitraty number. Async in a new thread is easier to test than async on _maybe_ new thread.

tciesla-atlassian (approved):

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -0,0 +1,75 @@
...
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+                return value;
+            });
+        });
+        Thread.sleep(10);
```
wyrzyk, line 69: Why do we need this sleep?

wyrzyk: https://github.com/atlassian-labs/db-replica/pull/71

PR #74, opened 2021-04-09
Make `LazyReference` thread-safe

wyrzyk:


wyrzyk:
Zrzut ekranu 2021-04-9 o 10 56 47

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
```diff
@@ -1,31 +1,36 @@
 package com.atlassian.db.replica.internal;
 
 
-import java.util.concurrent.atomic.AtomicReference;
+import com.atlassian.db.replica.internal.util.ThreadSafe;
+
 import java.util.function.Supplier;
 
+@ThreadSafe
 public abstract class LazyReference<T> implements Supplier<T> {
```
wyrzyk, line 9: - [ ]  hide the change with `compatibleWithPreviousVersion`.

wyrzyk:
The connection leak can occur when we share a connection between two threads and query the database simultaneously. Postgres do not support this, so we shouldn't be affected. Anyway, it's dangerous to use non-thread-safe Lazy reference, and I'd like to fix it.

src/test/java/com/atlassian/db/replica/internal/LazyReferenceTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        countingReference.get();
+        countingReference.get();
+        countingReference.get();
+
+        assertThat(countingReference.getCounter()).isEqualTo(1);
+    }
+
+    @Test
+    public void shouldCreateValueOnceWhileAccessedConcurrently() throws InterruptedException {
```
dagguh, line 27: Really nice test

src/test/java/com/atlassian/db/replica/internal/LazyReferenceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private void fetchReference(CountingReference countingReference, CountDownLatch start, CountDownLatch end) {
+        start.countDown();
+        try {
+            start.await();
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+        countingReference.get();
+        end.countDown();
+    }
```
dagguh, line 51: We can get a cleaner synchronization code if we turn the `Runnable` into a `Callable`
```suggestion
    private int fetchReference(
        LazyReference<Integer> reference,
        CountDownLatch start,
        CountDownLatch end
    ) throws InterruptedException {
        start.countDown();
        start.await();
        int value = reference.get();
        end.countDown();
        return value;
    }
```
If we don't want to rely on `Integer`, then we can turn `void` into `Void` and we'll be good too.

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
```diff
@@ -1,31 +1,36 @@
...
     public boolean isInitialized() {
-        return reference.get() != null;
+        return reference != null;
     }
 
     @Override
     public T get() {
-        if (!isInitialized()) {
-            try {
-                reference.compareAndSet(null, create());
```
dagguh, line 20: This CAS should have worked earlier too. If it returns `false`, then another thread was able to `create`, so we're good.
I ran `git checkout upstream/master -- src/main` and the test is still green.

- [ ] make sure the change is needed

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
wyrzyk: It's strange. The CAS should not work as create call is not protected by CAS. Also, the test was red for me, and I call it multiple times. The test also failed on CI (see the screenshot).
1. CAS will help us not to set the reference twice.
2. CAS will not protect from allocating a connection.

dagguh: You're right. Maybe I was lucky

wyrzyk: It's hard to write a concurrency test to reproduce a race condition that is not flaky and doesn't run forever 😢 .

dagguh (approved):

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
dagguh: No worries 🙂 
It's still great that it reproduced a few times.

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
dagguh: Yep, let's hide it

dagguh (changes requested):

src/test/java/com/atlassian/db/replica/internal/LazyReferenceTest.java
afaruga-atlassian, line 27: Good test

afaruga-atlassian:
Once hide I'm happy to approve.

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
dwatl: @wyrzyk I think it's done, right?

dwatl (approved):

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
```diff
@@ -1,31 +1,78 @@
...
     }
 
-    public void reset() {
-        reference.set(null);
+    private void reset_old() {
+        atomicReference.set(null);
+    }
+
+    private T get_new() {
+        synchronized (lock) {
```
afaruga-atlassian, line 59: So you pivot from `atomicReference` to lock with `object`

dagguh (approved):
I cannot review this anymore, so I'll just unblock you

afaruga-atlassian (approved):

afaruga-atlassian:
>  cannot review this anymore, so I'll just unblock you

👍 
Damn, I'm reviewing merged stuff. :trollface:

wyrzyk:
@afaruga-atlassian There's nothing wrong with reviewing merged stuff.

PR #75, opened 2021-04-13
Support `SELECT FOR UPDATE SKIP LOCKED` statement

wyrzyk:
I initially started with sql.contains without trimming. This implementation
didn't pass the test. Added some trimming to focus only on the end of the
query if the query is long.


tciesla-atlassian (approved):
LGTM

dagguh:
Red checks:
* https://github.com/atlassian-labs/db-replica/runs/2340847527
* https://github.com/atlassian-labs/db-replica/actions/runs/751968279

> java.lang.AssertionError: [thruput per ms] 
Expecting:
 <2310.3225f>
to be greater than:
 <2500.0f> 

> java.lang.AssertionError: [thruput per ms] 
Expecting:
 <2371.8604f>
to be greater than:
 <2500.0f> 

I remember times when the expected thruput was 5000.
The cost of features 😅
It's still high enough. I'll just peg it down to 2200.

wyrzyk:
@dagguh, I think in the end, we can't support 100% of cases without a large performance penalty. We may consider exposing the checks via API/SPI and allow customers to write their own checks or skip some expensive ones.

The other more crazy idea would handle the issue itself (catch `cannot execute X in a read-only transaction`, re-run on the main database and add to an `always main` list).

dagguh:
In relative terms, we got 2x worse thruput than a few versions earlier. But in absolute terms, it's still way more than necessary. Routing 2000 calls to replica at a cost of a millisecond is a good deal.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -648,7 +657,8 @@ public ReplicaStatement build() {
                 resultSetType,
                 resultSetConcurrency,
                 resultSetHoldability,
-                readOnlyFunctions
+                readOnlyFunctions,
+                compatibleWithPreviousVersion
```
afaruga-atlassian, line 661: I've found myself scrolling fast through the boiler code related to:`compatibleWithPreviousVersion`
It might be good idea (for review purposes) to extract separate commit - adding `compatibleWithPreviousVersion`

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -27,4 +43,12 @@ private boolean isUpdate() {
     private boolean isDelete() {
         return sql != null && (sql.startsWith("delete") || sql.startsWith("DELETE"));
     }
+
+    private String trimForSelectForUpdateCheck() {
+        if (sql.length() < SELECT_FOR_UPDATE_SUFFIX_LIMIT) {
+            return sql;
+        } else {
+            return sql.substring(sql.length() - SELECT_FOR_UPDATE_SUFFIX_LIMIT);
```
afaruga-atlassian, line 51: considering that they still will "SELECT FOR UPDATES" hidden in 100+ chars, (very low likelihood) how they can be noticed?

PR #79, opened 2021-04-16
Implemented missing methods in `Statement`

wyrzyk:


src/test/java/com/atlassian/db/replica/it/ReplicaStatementIT.java
```diff
@@ -0,0 +1,120 @@
...
+            statement.setLargeMaxRows(12345);
+            assertThatThrownBy(statement::getLargeMaxRows).isInstanceOf(SQLFeatureNotSupportedException.class);
+            assertThatThrownBy(statement::executeLargeBatch).isInstanceOf(SQLFeatureNotSupportedException.class);
+            assertThatThrownBy(() -> statement.executeLargeUpdate("CREATE SEQUENCE mysequence2;"))
+                .isInstanceOf(SQLFeatureNotSupportedException.class);
+            assertThatThrownBy(() -> statement.executeLargeUpdate("CREATE SEQUENCE mysequence2;", new int[]{}))
+                .isInstanceOf(SQLFeatureNotSupportedException.class);
+            assertThatThrownBy(() -> statement.executeLargeUpdate("CREATE SEQUENCE mysequence2;", new String[]{}))
+                .isInstanceOf(SQLFeatureNotSupportedException.class);
+            assertThatThrownBy(statement::cancel).isInstanceOf(SQLFeatureNotSupportedException.class);
```
dagguh, line 116: So this is like an executable to do list. Nice!

dagguh (approved):

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -4,24 +4,54 @@
     private static final int SELECT_FOR_UPDATE_SUFFIX_LIMIT = 100;
 
     private final String sql;
+    private final boolean compatibleWithPreviousVersion;
 
-    public SqlQuery(String sql) {
+    public SqlQuery(String sql, boolean compatibleWithPreviousVersion) {
```
dagguh, line 9: So `sql` is nullable? When?
We should consider `@Nullable`.

wyrzyk: We pass it from RouteDecisionBuilder builder.
Zrzut ekranu 2021-04-19 o 12 00 41

dagguh: `new SqlQuery(decisionBuilder.getSql())` 
A `SqlQuery`, which doesn't have the query, doesn't make sense in the domain. 🤔 It shouldn't be created in case of `decisionBuilder.getSql() == null`.

dagguh: https://github.com/atlassian-labs/db-replica/pull/81

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -1,43 +1,16 @@
...
-    private static final int SELECT_FOR_UPDATE_SUFFIX_LIMIT = 100;
-
-    private final String sql;
-
-    public SqlQuery(String sql) {
-        if (sql == null) {
-            throw new RuntimeException("An SqlQuery must have an SQL query string");
+public interface SqlQuery {
+    static SqlQuery create(String sql, boolean compatibleWithPreviousVersion){
+        if(compatibleWithPreviousVersion){
```
dagguh, line 5: A few `if`s replaced by a single `if` + temporary polymorphism. Seems like an acceptable tradeoff

dagguh (approved):

PR #81, opened 2021-04-19
Require an SQL query in `SqlQuery`

dagguh:


wyrzyk (approved):

dagguh:
Tests from https://github.com/atlassian-labs/db-replica/pull/79 cover this. @wyrzyk wanna cherry-pick this into #79 add tests here or merge as is?

wyrzyk:
@dagguh The test doesn't cover it. The test asserts it's safe to use null for SqlQuery, and this implementation will throw a RuntimeException.

dagguh:
`ReplicaStatementIT` covers it by calling `statement.executeQuery("SELECT 1;");`. I didn't add that test here to avoid conflicts.

dagguh:
PS. It seems that in all cases the `String sql` is already not null. And `getReadStatement` + `getWriteStatement` + `getCurrentStatement` should become `protected` at most.

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -558,15 +558,17 @@ public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) {
...
-        if (sqlQuery.isWriteOperation(sqlFunction)) {
-            decisionBuilder.reason(WRITE_OPERATION);
-            return prepareWriteStatement(decisionBuilder);
-        }
-
-        if (sqlQuery.isSelectForUpdate()) {
-            decisionBuilder.reason(LOCK);
-            return prepareWriteStatement(decisionBuilder);
+        String sql = decisionBuilder.getSql();
+        if (sql != null) {
```
dagguh, line 562: `ReplicaStatementIT` covers triggers the new `if` by calling `statement.executeQuery("SELECT 1;");` in #79 . I didn't add that test here to avoid conflicts.

dagguh: The only way to make this `null` is to call `getReadStatement` directly. But it's only called internally. It could be `protected` (maybe even `private`). I don't see a reason for it being public. PS. `getWriteStatement` + `getCurrentStatement` could also become `protected` at most.

wyrzyk: >ReplicaStatementIT covers triggers the new if by calling statement.executeQuery("SELECT 1;");

Yes. But for this case, SQL != null so that it will be green before and after.

PR #83, opened 2021-04-21
Add `preWrite` to `ReplicaConsistency`

wyrzyk:
Internally we are experimenting with a new way of tracking inconsistencies. We
increase an integer in the transaction just after changes committed. We use
`ReplicaConsistency#write`. This approach has some disadvantages:
- it doubles the number of transactions. It also doubles the number of changes to be replicated. It may affect replication lag.
- we need two database calls to propagate the information (update + commit).
- We increase `inconsistency` time. We don't track when changes are propagated, but we handle changes after the changes are propagated.

`Connection#preWrite` allows us to inject our counter bump into the same transaction.

### DIFF
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
@@ -24,12 +24,15 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     private final ConnectionState state;
     private final ConnectionParameters parameters;
     private final Warnings warnings;
+    private final boolean compatibleWithPreviousVersion;
 
     public ReplicaConnectionProvider(
         ConnectionProvider connectionProvider,
         ReplicaConsistency consistency,
-        StateListener stateListener
+        StateListener stateListener,
+        boolean compatibleWithPreviousVersion
     ) {
+        this.compatibleWithPreviousVersion = compatibleWithPreviousVersion;
         this.parameters = new ConnectionParameters();
         this.warnings = new Warnings();
         this.state = new ConnectionState(connectionProvider, consistency, parameters, warnings, stateListener);
@@ -58,6 +61,11 @@ public class ReplicaConnectionProvider implements AutoCloseable {
 
     public void setAutoCommit(Boolean autoCommit) throws SQLException {
         final boolean autoCommitBefore = getAutoCommit();
+        if (!compatibleWithPreviousVersion) {
+            if (autoCommitBefore != autoCommit) {
+                preCommit(autoCommitBefore);
+            }
+        }
         parameters.setAutoCommit(state::getConnection, autoCommit);
         if (autoCommitBefore != getAutoCommit()) {
             recordCommit(autoCommitBefore);
@@ -168,6 +176,9 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     public void commit() throws SQLException {
         final Optional<Connection> connection = state.getConnection();
         if (connection.isPresent()) {
+            if (!compatibleWithPreviousVersion) {
+                preCommit(parameters.isAutoCommit());
+            }
             connection.get().commit();
             recordCommit(parameters.isAutoCommit());
         }
@@ -175,7 +186,15 @@ public class ReplicaConnectionProvider implements AutoCloseable {
 
     private void recordCommit(boolean autoCommit) throws SQLException {
         if (state.getState().equals(MAIN) && !autoCommit) {
-            consistency.write(state.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL)));
+            final Connection mainConnection = state.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL));
+            consistency.write(mainConnection);
+        }
+    }
+
+    private void preCommit(boolean autoCommit) throws SQLException {
+        if (state.getState().equals(MAIN) && !autoCommit) {
+            final Connection mainConnection = state.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL));
+            consistency.preCommit(mainConnection);
         }
     }
 
--- a/src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
+++ b/src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
@@ -12,11 +12,23 @@ import java.util.function.Supplier;
 public interface ReplicaConsistency {
 
     /**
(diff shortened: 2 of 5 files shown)


wyrzyk:
@tciesla-atlassian we can use it to improve `sequence-value-consistency-mode`. It may reduce the replication lag.

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -58,6 +58,9 @@ public int getTransactionIsolation() throws SQLException {
 
     public void setAutoCommit(Boolean autoCommit) throws SQLException {
         final boolean autoCommitBefore = getAutoCommit();
+        if (autoCommitBefore != autoCommit) {
```
wyrzyk, line 61: It's a draft because I didn't add `compatibleWithPreviousVersion`. I want to get the change reviewed first and then add `ifs`.

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
```diff
@@ -12,12 +12,24 @@
...
 
+    /**
+     * Invoked just before transaction commit.
+     *
+     * Notice: The method will not handle all writes. Writes done outside of a transaction
+     * needs to be handled in `ReplicaConnection#write`.
+     *
+     * @param main connects to the main database
+     */
+    default void preWrite(Connection main){
```
dagguh, line 30: Thanks for the `default`. Source compatibility maintained 😌

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
afaruga-atlassian: What're your thoughts on adding those `ifs` every time in - **crucial place**, and boiler code in all top-down classes to provide if in the **crucial place**
I remember that at the beginning I predict that there was the conclusion to avoid any static providers.
I just want to ensure you still think so.

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
```diff
@@ -12,12 +12,24 @@
...
+     * when in a transaction.
      *
      * @param main connects to the main database
      */
     void write(Connection main);
 
+    /**
+     * Invoked just before transaction commit.
+     *
+     * Notice: The method will not handle all writes. Writes done outside of a transaction
```
afaruga-atlassian, line 25: > The method will not handle all writes

have you thought about adding a test/any other check to check which will be handled, which not?

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
afaruga-atlassian, line 30: Good that this is added in `SPI`, it would be risky in `API`

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
wyrzyk: > What're your thoughts on adding those ifs every time in - crucial place

I'm happy that we don't have to accumulate it, and we can clear it every release. In the product part, we accumulated 16 feature flags so far, while here we have only one. 
Yesterday I spend 3h pairing on the product flags dependencies and trying to unsolve the puzzle. I prefer to add the code (you just add the if and compiler + intellij will tell you where should you add the boilerplate). It's much easier than managing complex dependencies.

> every time in - crucial place, and boiler code in all top-down classes to provide if in the crucial place

We should not be afraid to change critical places. If there's a path uncovered by tests, we'll introduce the bug anyway in the future, and we add the missing coverage.

>  was the conclusion to avoid any static providers. I just want to ensure you still think so.

Yes. We should avoid static providers. Nothing changed here.

src/test/java/com/atlassian/db/replica/api/TestConsistency.java
```diff
@@ -26,7 +27,11 @@ public void shouldRefreshCommitted() throws SQLException {
         connection.prepareStatement(SIMPLE_QUERY).executeUpdate();
         connection.commit();
 
-        verify(consistency).write(any());
+        final Connection mainConnection = connectionProvider.getProvidedConnections().get(0);
+        final InOrder inOrder = inOrder(consistency, mainConnection);
+        inOrder.verify(consistency).preWrite(any());
```
afaruga-atlassian, line 32: and everything visible in tests <3

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
wyrzyk: Why?

wyrzyk: Adding methods to SPI is risky. Adding methods to API is not risky. 😉

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
```diff
@@ -12,12 +12,24 @@
...
+     * needs to be handled in `ReplicaConnection#write`.
+     *
+     * @param main connects to the main database
+     */
+    default void preWrite(Connection main){
+    }
+
```
afaruga-atlassian, line 32: It there is a need for `preWrite` maybe there is a need for `afterWrite`
How about an interceptor?

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
wyrzyk: I thought about copy-pasting our current `ReplicaConsistency` implementation from the product and adding IT tests. I think I covered all the cases with unit tests.

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
wyrzyk: `afterWrite` = write. I thought about renaming it. We would have to deprecate `write`. I'm still not convinced if it's worth it. @dagguh WDYT?

dagguh: Clarity is worth a deprecation

wyrzyk: I guess it should be
`write` and `preCommit`. 
`afterWrite` and `preWrite` would be misleading. It would seem to be symmetrical, while it's not. We can hook before commit only if there's a transaction. `preWrite`/`preCommit` is a subset of `write`.

dagguh (approved):

PR #85, opened 2021-05-28
Allow to opt-out from replication

wyrzyk:
Scenarios when we want to opt-out from the replica:
- we have special contexts where we want to load data from
the main database only (to be 100% sure the data is always consistent, regardless of ReplicaConsistency implementations).
- limit the load that goes to the replica.
- turn off calls to a replica without waiting for DualConnections to be closed.

Until now, we were using `ReplicaConsistency#isConsistent` to guard
calls to a replica. It works, but it also is confusing (why replica is not consistent in this context?). The goal of
this change is to provide a dedicated SPI.


wyrzyk:
@dagguh I think it can help with the refactoring and help us to reduce the number of ReplicaConsistency implementations that have nothing in common with consistency.

dagguh:
There are many `Reason`s to avoid replica. Consistency is one of them. Perhaps a common SPI for `ReplicaAvoidance` could cover all those cases.

We might say "but limiting replica load doesn't need `Supplier<Connection> replica`. True, but so does `PessimisticPropagationConsistency` and many others. That's why it's a `Supplier`. So the generic SPI could also be provided with the "could be useful" params.

wyrzyk:
Yes, the new SPI is redundant. We can make ReplicaConsistency more generic.

wyrzyk:
I was not precise enough. I meant adding a supplementary SPI, and keeping ReplicaConsistency = we'll add a redundant SPI. Make ReplicaConsistency more generic -> I meant providing alternative more generic SPI and deprecating ReplicaConsistency. Those were my thought shortcuts, as from an API evolution point of view, we either add or change.

wyrzyk:
We will also lose `Reason` granularity unless the new SPI could control the decision, instead of just a boolean.

dagguh:
The new SPI should indeed control (`return`?) the `RouteDecision`. Perhaps it could also absorb APIs like `DualConnection.Builder.readOnlyFunctions`.

wyrzyk:
I started to rewrite our internals to a new `decision-making` SPI. @dagguh Do you think it's a good direction?

src/main/java/com/atlassian/db/replica/internal/ConsistencyGuardingChief.java
```diff
@@ -0,0 +1,29 @@
...
+    private final ReplicaConsistency replicaConsistency;
+
+    public ConsistencyGuardingChief(ReplicaConsistency replicaConsistency) {
+        this.replicaConsistency = replicaConsistency;
+    }
+
+    @Override
+    public void overrideDecision(
+        RouteDecisionBuilder decisionBuilder,
+        Collection<Supplier<Connection>> replicaConnections
```
dagguh, line 22: Future proofing for multi-replicas 👌

src/main/java/com/atlassian/db/replica/internal/ConsistencyToTransactionHookAdapter.java
```diff
@@ -0,0 +1,24 @@
+package com.atlassian.db.replica.internal;
+
+import com.atlassian.db.replica.spi.ReplicaConsistency;
+import com.atlassian.db.replica.spi.TransactionHook;
+
+import java.sql.Connection;
+
+public class ConsistencyToTransactionHookAdapter implements TransactionHook {
```
dagguh, line 8: So `ReplicaConsistency` doesn't implement `TransactionHook`, because we want to make `preCommit` explicit rather than implicitly no-op?

src/main/java/com/atlassian/db/replica/spi/ReplicaConsistency.java
```diff
@@ -7,8 +7,11 @@
 
 /**
  * Tracks data consistency between replica and main databases.
+ *
+ * @deprecated TODO
```
dagguh, line 11: - [ ] `@deprecated TODO`

src/main/java/com/atlassian/db/replica/spi/TransactionHook.java
```diff
@@ -0,0 +1,23 @@
+package com.atlassian.db.replica.spi;
+
+import java.sql.Connection;
+
+public interface TransactionHook {
+    /**
+     * Informs that {@code main} received an UPDATE, INSERT or DELETE or transaction commit
```
dagguh, line 7: Doesn't it also inform about `SELECT writingFunction()` and other JDBC non-execute API calls?

src/main/java/com/atlassian/db/replica/spi/TransactionHook.java
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * @param main connects to the main database
+     */
+    void write(Connection main);
+
+    /**
+     * Invoked just before transaction commit.
+     * <p>
+     * Notice: The method will not handle all writes. Writes done outside of a transaction
+     * needs to be handled in `TransactionHook#write`.
```
dagguh, line 18:
```suggestion
     * Invoked just before an explicit transaction commit.
     * <p>
     * Notice: The method will not handle all commits. Auto-commits need to be handled in `TransactionHook#write`.
```

src/main/java/com/atlassian/db/replica/spi/TransactionHook.java
```diff
@@ -0,0 +1,23 @@
...
+import java.sql.Connection;
+
+public interface TransactionHook {
+    /**
+     * Informs that {@code main} received an UPDATE, INSERT or DELETE or transaction commit
+     * when in a transaction.
+     *
+     * @param main connects to the main database
+     */
+    void write(Connection main);
```
dagguh, line 12: The past tense of "received an UPDATE" suggests this is called after the write.

```suggestion
    void postWrite(Connection main);
```

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -189,17 +189,15 @@ private Connection prepareReadConnection(RouteDecisionBuilder decisionBuilder) t
...
                 closeConnection(writeConnection, decisionBuilder);
             }
-            final Connection connection = readConnection.get(decisionBuilder);
-            replicaConsistent = true;
-            return connection;
-        } else {
-            replicaConsistent = false;
-            decisionBuilder.reason(REPLICA_INCONSISTENT);
-            return prepareMainConnection(decisionBuilder);
+            return readConnection.get(decisionBuilder);
```
dagguh, line 200: The `if` builds the decision, but here we're still passing the partial one. Is that intentional?

src/main/java/com/atlassian/db/replica/internal/ReplicaStatement.java
```diff
@@ -569,10 +563,6 @@ public Statement getReadStatement(RouteDecisionBuilder decisionBuilder) throws S
         String sql = decisionBuilder.getSql();
         if (sql != null) {
             SqlQuery sqlQuery = new SqlQuery(sql);
-            if (sqlQuery.isWriteOperation(sqlFunction)) {
```
dagguh, line 572: This check could disappear from the UML and be replaced by "default chief UML".

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -189,17 +189,15 @@ private Connection prepareReadConnection(RouteDecisionBuilder decisionBuilder) t
             decisionBuilder.cause(writeConnection.getFirstCause().build());
             return writeConnection.get(decisionBuilder);
         }
-        if (consistency.isConsistent(() -> readConnection.get(decisionBuilder))) {
```
dagguh, line 192: This `if` could be updated in the UML

src/main/java/com/atlassian/db/replica/spi/Chief.java
```diff
@@ -0,0 +1,14 @@
...
+import com.atlassian.db.replica.internal.RouteDecisionBuilder;
+
+import java.sql.Connection;
+import java.util.Collection;
+import java.util.function.Supplier;
+
+/**
+ * Can override the automatic decision to make a final call where the query should go.
+ */
+public interface Chief {
```
dagguh, line 12: Like a boss 😉

src/main/java/com/atlassian/db/replica/internal/DatabaseMicroManager.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.sql.Connection;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import static com.atlassian.db.replica.api.reason.Reason.WRITE_OPERATION;
+
+public class DatabaseMicroManager implements Chief {
```
dagguh, line 14: I like the micromanagement metaphor 😃 
But "database" seems too wide for a responsibility.

```suggestion
public class SqlQueryChief implements Chief {
```

src/main/java/com/atlassian/db/replica/spi/Chief.java
dagguh, line 12:
```suggestion
/**
 * Can reroute the current call by overriding the current decision.
 */
public interface Router {
```

Impls would go `SqlQueryChief` -> `SqlQueryRouter`, `ConsistencyGuardingChief` -> `ConsistencyRouter`.

src/main/java/com/atlassian/db/replica/internal/ConsistencyToTransactionHookAdapter.java
dagguh, line 8:
```suggestion
public class ConsistencyTransactionHook implements TransactionHook {
```

src/main/java/com/atlassian/db/replica/spi/Chief.java
```diff
@@ -0,0 +1,14 @@
+package com.atlassian.db.replica.spi;
+
+import com.atlassian.db.replica.internal.RouteDecisionBuilder;
+
+import java.sql.Connection;
+import java.util.Collection;
+import java.util.function.Supplier;
+
+/**
+ * Can override the automatic decision to make a final call where the query should go.
```
dagguh, line 10: Could we clarify the lifecycle of a `Chief`? Is it reused across multiple database calls? Multiple connections? Multiple pools?

We can use the [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) MUST, SHOULD, MAY keywords.

src/main/java/com/atlassian/db/replica/api/DualConnection.java
```diff
@@ -59,14 +64,18 @@ private DualConnection(
         Set<String> readOnlyFunctions,
         boolean compatibleWithPreviousVersion
     ) {
+        final ChainOfChiefs chiefs = new ChainOfChiefs(Arrays.asList(
```
dagguh, line 67: Gotta open a way to inject custom `Chief`s.

- [ ] inject custom `Chief`

dagguh (changes requested):

src/main/java/com/atlassian/db/replica/internal/ChainOfChiefs.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public void overrideDecision(
+        RouteDecisionBuilder decisionBuilder,
+        Collection<Supplier<Connection>> replicaConnections
+    ) {
+        for (Chief chief : chiefs) {
+            chief.overrideDecision(decisionBuilder, replicaConnections);
+            if (decisionBuilder.build().willRunOnMain()) {
```
wyrzyk, line 24: I should not be here. Chefs should have more control.

src/main/java/com/atlassian/db/replica/internal/ConsistencyToTransactionHookAdapter.java
wyrzyk: Yes. No-op was there not to break the API.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: I need to verify it.

src/main/java/com/atlassian/db/replica/spi/Chief.java
wyrzyk: The life cycle is up to the customer. For example, a customer can inject a new instance every DualConnection or use a singleton.

src/main/java/com/atlassian/db/replica/spi/Chief.java
wyrzyk: I still hope someone will find a better name 🤞.

src/main/java/com/atlassian/db/replica/spi/Chief.java
dagguh: MAY the customer choose to have a `new` Chief for every `DatabaseCall` or MUST every Chief impl handle multiple calls?
Is `Chief` lifecycle equivalent to `DualConnection` lifecycle? Does it have to be `@ThreadSafe` or is it ok to be `@NotThreadSafe`?
If reusing `Chief`s across DB pool is recommended, then they SHOULD do it. If it's recommended to avoid that, then they SHOULD NOT do it. If it's ok either way, they MAY do it. OTOH the stance on this lifecycle probably SHOULD be documented in the `DuallConnection.Builder` injection site rather than in the `Chief` class itself.

src/main/java/com/atlassian/db/replica/internal/ChainOfChiefs.java
afaruga-atlassian: ++ 
> if (decisionBuilder.build().willRunOnMain())

reduce cohesivness of this class.

src/main/java/com/atlassian/db/replica/internal/ConsistencyGuardingChief.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ConsistencyGuardingChief implements Chief {
+    private final ReplicaConsistency replicaConsistency;
+
+    public ConsistencyGuardingChief(ReplicaConsistency replicaConsistency) {
+        this.replicaConsistency = replicaConsistency;
+    }
+
+    @Override
+    public void overrideDecision(
```
afaruga-atlassian, line 20: It looks strange. This gives a feeling that MultiReplicas aren't supported, and their behaviour needs to be overridden. 
I thought `Chief` should be used to override default decision.

I would rather add `withMultipleReplicas(Builder)` method to `DualConnection.Builder` to add support for them. WDYT?

afaruga-atlassian (changes requested):

PR #86, opened 2021-05-31
Clarify `ReplicaConsistency` vs commits

dagguh:
This is how I understood the SPI. This draft is just a place for discussion and clarification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.3.0...master
 
+### Added
+- Call `ReplicaConsistency.preCommit` only for explicit commits.
+- Clarify the relationship of `ReplicaConsistency.write` and auto-commit.
+
 ## [1.3.0] - 2021-04-21
 [1.3.0]: https://github.com/atlassian-labs/db-replica/compare/release-1.2.6...release-1.3.0
 
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
@@ -57,14 +57,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
     }
 
     public void setAutoCommit(Boolean autoCommit) throws SQLException {
-        final boolean autoCommitBefore = getAutoCommit();
-        if (autoCommitBefore != autoCommit) {
-            preCommit(autoCommitBefore);
-        }
         parameters.setAutoCommit(state::getConnection, autoCommit);
-        if (autoCommitBefore != getAutoCommit()) {
-            recordCommit(autoCommitBefore);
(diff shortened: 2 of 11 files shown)


src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -57,14 +57,7 @@ public int getTransactionIsolation() throws SQLException {
     }
 
     public void setAutoCommit(Boolean autoCommit) throws SQLException {
-        final boolean autoCommitBefore = getAutoCommit();
-        if (autoCommitBefore != autoCommit) {
-            preCommit(autoCommitBefore);
-        }
         parameters.setAutoCommit(state::getConnection, autoCommit);
-        if (autoCommitBefore != getAutoCommit()) {
-            recordCommit(autoCommitBefore);
```
dagguh, line 66: Why did `ReplicaConsistency` need to know about `setAutoCommit`? Can this call cause an immediate commit?
I thought that the next `executeUpdate` could do that implicitly, but nothing happens in `setAutoCommit` itself.

dagguh: I added the integration tests. One fails when introduced initially and goes green after removing auto-commit handling.

src/test/java/com/atlassian/db/replica/api/TestConsistency.java
```diff
@@ -137,22 +136,6 @@ public void shouldNotRefreshAfterSelectForUpdate() throws SQLException {
         verify(consistency, never()).write(any());
     }
 
-    @Test
-    public void shouldSetAutoCommitRefreshWhenInTransaction() throws SQLException {
```
afaruga-atlassian, line 141: If  test is deleted, shouldn't a new one be created for reusing transactions 🤔

dagguh: There are two new tests in `DualConnectionIT`.

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.3.0...master
 
+### Added
+- Call `ReplicaConsistency.preCommit` only for explicit commits.
```
wyrzyk, line 28: It was working this way from the beginning.

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
wyrzyk: > Why did ReplicaConsistency need to know about setAutoCommit? Can this call cause an immediate commit?

Yes. It's a part of the connection API. 
```
   * <B>NOTE:</B>  If this method is called during a transaction and the
     * auto-commit mode is changed, the transaction is committed.  If
     * <code>setAutoCommit</code> is called and the auto-commit mode is
     * not changed, the call is a no-op.
```

dagguh:
Thanks for clarifying

PR #88, opened 2021-06-14
Replicas discovery

dwatl:
AuroraReplicasDiscoverer makes is easy to fetch direct urls to all replicas. It could be handy when you don't want to depend on default AWS load balancing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.3.0...master
 
+### Added
+- `AuroraReplicasDiscoverer`
+
 ## [1.3.0] - 2021-04-21
 [1.3.0]: https://github.com/atlassian-labs/db-replica/compare/release-1.2.6...release-1.3.0
 
--- a/src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
+++ b/src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
@@ -1,0 +1,55 @@
+package com.atlassian.db.replica.api.aurora;
+
+import com.atlassian.db.replica.internal.aurora.AuroraEndpoint;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.LinkedList;
+import java.util.List;
+
+import static com.atlassian.db.replica.internal.aurora.AuroraEndpoints.instanceEndpoint;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * It allows to pull Aurora Replicas cluster information
+ */
+public final class AuroraReplicasDiscoverer {
+    private final Connection connection;
+
+    public AuroraReplicasDiscoverer(Connection connection) {
+        this.connection = connection;
+    }
+
+    /**
+     * It provides list of replica endpoints based on reader endpoint and server ids
+     * @param readerEndpoint - Aurora reader endpoint
+     * @return list of replicas endpoints
+     * @throws SQLException
+     */
+    public List<AuroraEndpoint> fetchReplicasEndpoints(String readerEndpoint) throws SQLException {
+        return fetchReplicasServerIds()
+            .stream()
+            .map(serverId -> instanceEndpoint(readerEndpoint, serverId))
+            .collect(toList());
+    }
+
+    /**
+     * Asks Aurora about the list of server ids
+     * @return server ids
+     * @throws SQLException
+     */
+    public List<String> fetchReplicasServerIds() throws SQLException {
+        LinkedList<String> ids = new LinkedList<>();
+        try (ResultSet rs =
+                 connection.prepareStatement(
+                     "SELECT server_id FROM aurora_global_db_instance_status() WHERE session_id != 'MASTER_SESSION_ID'")
+                     .executeQuery()) {
+            while (rs.next()) {
+                ids.add(rs.getString("server_id"));
+            }
+        }
+        return ids;
+    }
+}
+
--- a/src/main/java/com/atlassian/db/replica/internal/aurora/AuroraCluster.java
+++ b/src/main/java/com/atlassian/db/replica/internal/aurora/AuroraCluster.java
@@ -1,0 +1,100 @@
+package com.atlassian.db.replica.internal.aurora;
+
+import java.util.Objects;
+
+import static com.atlassian.db.replica.internal.aurora.AuroraCluster.AuroraClusterBuilder.anAuroraCluster;
+
+public final class AuroraCluster {
+    private final String clusterName;
+    private final String clusterPrefix;
+
+    private AuroraCluster(String clusterName, String clusterPrefix) {
+        this.clusterName = clusterName;
+        this.clusterPrefix = clusterPrefix;
+    }
+
+    public static AuroraCluster parse(String cluster) {
+        Objects.requireNonNull(cluster);
+
+        String[] chunks = splitByLastOccurence("-", cluster);
+        if (chunks.length == 1) {
+            String clusterName = chunks[0];
+            return anAuroraCluster(clusterName).build();
+        } else {
+            String clusterName = chunks[0];
+            String clusterPrefix = chunks[1];
+            return anAuroraCluster(clusterName).clusterPrefix(clusterPrefix).build();
+        }
+    }
+
+    private static String[] splitByLastOccurence(String delimiter, String str) {
+        int i = str.lastIndexOf(delimiter);
+        if (i == -1) {
+            return new String[]{str};
+        } else {
+            return new String[]{str.substring(i + 1), str.substring(0, i)};
+        }
+    }
+
+    @Override
+    public String toString() {
+        if (clusterPrefix != null && !clusterPrefix.isEmpty()) {
+            return String.format("%s-%s", clusterPrefix, clusterName);
+        } else {
+            return clusterName;
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        AuroraCluster that = (AuroraCluster) o;
+
+        if (!Objects.equals(clusterName, that.clusterName)) return false;
+        return Objects.equals(clusterPrefix, that.clusterPrefix);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = clusterName != null ? clusterName.hashCode() : 0;
+        result = 31 * result + (clusterPrefix != null ? clusterPrefix.hashCode() : 0);
+        return result;
+    }
+
+    public String getClusterName() {
+        return clusterName;
+    }
+
+    public String getClusterPrefix() {
+        return clusterPrefix;
+    }
+
+
+    public static final class AuroraClusterBuilder {
+        private String clusterName;
+        private String clusterPrefix;
+
+        private AuroraClusterBuilder() {
+        }
+
+        public static AuroraClusterBuilder anAuroraCluster(String clusterName) {
+            return new AuroraClusterBuilder().clusterName(clusterName);
+        }
+
+        public AuroraClusterBuilder clusterName(String clusterName) {
+            this.clusterName = clusterName;
+            return this;
+        }
+
+        public AuroraClusterBuilder clusterPrefix(String clusterPrefix) {
+            this.clusterPrefix = clusterPrefix;
+            return this;
+        }
+
+        public AuroraCluster build() {
+            return new AuroraCluster(clusterName, clusterPrefix);
+        }
+    }
+}
--- a/src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoints.java
+++ b/src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoints.java
@@ -1,0 +1,20 @@
+package com.atlassian.db.replica.internal.aurora;
+
+import static com.atlassian.db.replica.internal.aurora.AuroraCluster.AuroraClusterBuilder.anAuroraCluster;
+import static com.atlassian.db.replica.internal.aurora.AuroraEndpoint.AuroraEndpointBuilder.anAuroraEndpoint;
+
+public class AuroraEndpoints {
+    private AuroraEndpoints() {
+    }
+
+    /**
+     * Transforms reader endpoint to instance endpoint
+     */
+    public static AuroraEndpoint instanceEndpoint(String readerEndpoint, String serverId) {
+        AuroraEndpoint endpoint = AuroraEndpoint.parse(readerEndpoint);
+        return anAuroraEndpoint(endpoint)
+            .serverId(serverId)
+            .cluster(anAuroraCluster(endpoint.getCluster().getClusterName()).clusterPrefix(null).build())
+            .build();
+    }
+}
--- a/src/main/java/com/atlassian/db/replica/internal/aurora/RdsDns.java
+++ b/src/main/java/com/atlassian/db/replica/internal/aurora/RdsDns.java
@@ -1,0 +1,83 @@
+package com.atlassian.db.replica.internal.aurora;
+
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class RdsDns {
+    private static final Pattern DNS_PATTERN = Pattern.compile("([^.]+).([^:]*):?([0-9]+)?");
+    private final String region;
+    private final String domain;
+    private final Integer port;
+
+    public RdsDns(String region, String domain, Integer port) {
+        this.region = region;
+        this.domain = domain;
+        this.port = port;
+    }
+
+    public static RdsDns parse(String dns) {
+        Objects.requireNonNull(dns);
+
+        Matcher matcher = DNS_PATTERN.matcher(dns);
+        matcher.matches();
+        return new RdsDns(matcher.group(1), matcher.group(2), parseNullableInteger(groupOrNull(matcher, 3)));
+    }
+
+    private static String groupOrNull(Matcher matcher, int n) {
+        try {
+            return matcher.group(n);
+        } catch (IndexOutOfBoundsException e) {
+            return null;
+        }
+    }
+
+    private static Integer parseNullableInteger(String str) {
+        if (str == null) {
+            return null;
+        } else {
+            return Integer.valueOf(str);
+        }
+    }
+
+    public String getRegion() {
+        return region;
+    }
+
+    public String getDomain() {
+        return domain;
+    }
+
+    public Integer getPort() {
+        return port;
+    }
+
+    @Override
+    public String toString() {
+        if (port != null) {
+            return String.format("%s.%s:%s", region, domain, port);
+        } else {
+            return String.format("%s.%s", region, domain);
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        RdsDns rdsDns = (RdsDns) o;
+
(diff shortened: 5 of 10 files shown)


src/test/java/com/atlassian/db/replica/api/aurora/RdsDnsTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+class RdsDnsTest {
+
+    @Test
+    void fromString() {
+        RdsDns dns = RdsDns.fromString("us-east-1.rds.amazonaws.com:3306");
```
dwatl, line 12: Port should be optional

- [ ] test without port

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,55 @@
...
+    private final Connection connection;
+
+    public AuroraReplicasDiscoverer(Connection connection) {
+        this.connection = connection;
+    }
+
+    /**
+     * It provides list of replica urls based on reader endpoint and server ids
+     * @param readerEndpoint - Aurora reader endpoint
+     * @return list of replicas jdbcUrls
```
dagguh, line 25: Perhaps it should return objects rather than Strings? It would be a form of fail fast. Note that JDBC URLs aren't valid `URI`s directly due to the prefix, but it's ok: 

```java
    @Test
    public void shouldParseJdbcUrl() {
        JdbcUrl parsed = JdbcUrl.parse("jdbc:postgresql://x.y.z:5432/db?x=y");

        assertThat(parsed.getUrl().getHost()).isEqualTo("x.y.z");
    }

    @Test
    public void shouldReParseJdbcUrl() {
        String raw = "jdbc:postgresql://x.y.z:5432/db?x=y";

        Throwable thrown = catchThrowable(() -> JdbcUrl.parse(JdbcUrl.parse(raw).toString()));

        assertThat(thrown).doesNotThrowAnyException();
    }

    class JdbcUrl {

        private static final String PREFIX = "jdbc:";

        private final URI url;

        private JdbcUrl(URI url) {
            this.url = url;
        }

        public URI getUrl() {
            return url;
        }

        public static JdbcUrl parse(String rawJdbcUrl) {
            if (rawJdbcUrl.startsWith(PREFIX)) {
                URI parsed = URI.create(rawJdbcUrl.replaceFirst(PREFIX, ""));
                return new JdbcUrl(parsed);
            } else {
                String parseError = String.format("Cannot parse %s due to missing prefix '%s'", rawJdbcUrl, PREFIX);
                throw new IllegalArgumentException(parseError);
            }
        }

        @Override
        public String toString() {
            return PREFIX + url.toString();
        }
    }
```

PS. we got a very similar class in the consumer of this lib.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,55 @@
...
+     * @throws SQLException
+     */
+    public List<String> fetchReplicasUrls(String readerEndpoint) throws SQLException {
+        return fetchReplicaServerIds()
+            .stream()
+            .map(serverId -> replicaUrl(
+                readerEndpoint,
+                serverId
+            )).collect(Collectors.toList());
```
dagguh, line 34: I cannot resist fixing ad hoc chopping 😝 
```suggestion
            .map(serverId -> replicaUrl(readerEndpoint, serverId))
            .collect(toList());
```

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Asks Aurora about the list of server ids
+     * @return server ids
+     * @throws SQLException
+     */
+    public List<String> fetchReplicaServerIds() throws SQLException {
+        LinkedList<String> ids = new LinkedList<>();
+        try (ResultSet rs =
+                 connection.prepareStatement(
+                     "SELECT server_id FROM aurora_global_db_instance_status() WHERE session_id != 'MASTER_SESSION_ID'")
```
dagguh, line 46: Is it possible to integration test it in Docker?

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,55 @@
...
+        return fetchReplicaServerIds()
+            .stream()
+            .map(serverId -> replicaUrl(
+                readerEndpoint,
+                serverId
+            )).collect(Collectors.toList());
+    }
+
+    /**
+     * Asks Aurora about the list of server ids
```
dagguh, line 38: It seems like a behavior of `AuroraCluster`. It could ask for the `Connection` via method param.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: We don't have Aurora, we could create a function in postgres that will return the same rows all the time. Not ideal, but better than zero coverage, right?

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: It depends, I was looking at them as pure data classes. They simply transfer data and are not coupled with behavior.
However, I'm biased here so happy to consider pros and cons.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: Now that I think about it... It's not JdbcUrl. It's `instance endpoint` according to [AWS definitions](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Overview.Endpoints.html#Aurora.Overview.Endpoints.Types) (it has nothing to do with Jdbc).

src/main/java/com/atlassian/db/replica/api/aurora/AuroraCluster.java
```diff
@@ -0,0 +1,70 @@
...
+            return new AuroraCluster(chunks[0], chunks[1]);
+        }
+    }
+
+    private static String[] splitByLastOccurence(String delimiter, String str) {
+        int i = str.lastIndexOf(delimiter);
+        if (i == -1) {
+            return new String[]{str};
+        } else {
+            return new String[]{str.substring(i + 1), str.substring(0, i)};
```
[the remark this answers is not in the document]
dwatl: AuroraCluster has a cluster name and can have a prefix. I agree that local variables could unhide some of the magic here.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: I'm also good with using `parse` instead of `fromString` 👌 It's more Java-ish

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dagguh: Pure data classes = procedural style

src/main/java/com/atlassian/db/replica/api/aurora/AuroraCluster.java
```diff
@@ -0,0 +1,70 @@
+package com.atlassian.db.replica.api.aurora;
+
+import java.util.Objects;
+
+public class AuroraCluster {
```
wyrzyk, line 5: final

src/main/java/com/atlassian/db/replica/api/aurora/AuroraCluster.java
```diff
@@ -0,0 +1,70 @@
...
+    private final String clusterName;
+    private final String clusterPrefix;
+
+    public AuroraCluster(String clusterName) {
```
wyrzyk, line 9: Did you consider a builder?

wyrzyk: Or making constructors private (you are providing a factory method).

src/main/java/com/atlassian/db/replica/api/aurora/AuroraCluster.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public AuroraCluster(String clusterName, String clusterPrefix) {
+        this.clusterName = clusterName;
+        this.clusterPrefix = clusterPrefix;
+    }
+
+    public static AuroraCluster fromString(String cluster) {
+        String[] chunks = splitByLastOccurence("-", cluster);
+        if (chunks.length == 1) {
```
wyrzyk, line 20: No need to worry about cluster==null?

src/main/java/com/atlassian/db/replica/api/aurora/AuroraEndpoint.java
```diff
@@ -0,0 +1,84 @@
+package com.atlassian.db.replica.api.aurora;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class AuroraEndpoint {
```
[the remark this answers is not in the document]
wyrzyk: And please make all the API classes final to allow evolution in the future.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final Pattern ENDPOINT_PATTERN = Pattern.compile("([^.]+).([^.]+).(.*)");
+    private final String serverId;
+    private final AuroraCluster cluster;
+    private final RdsDns dns;
+
+    public AuroraEndpoint(String serverId, AuroraCluster cluster, RdsDns dns) {
```
wyrzyk, line 12: Why is it public? Builder is not enough to create the class?

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,55 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.atlassian.db.replica.api.aurora.AuroraEndpoints.replicaUrl;
+
+/**
+ * It allows to pull Aurora Replicas cluster information
```
wyrzyk, line 13: Why is it needed?

dwatl: Once you have instance endpoints you can build connection pool per replica.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraCluster.java
dwatl: Not sure, everything can be null in Java :/

dagguh: Not after a null check 😉

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: Not sure, these are encouraged everywhere lately, even Java is adding them "natively" (Java Records).

src/main/java/com/atlassian/db/replica/api/aurora/AuroraCluster.java
dwatl: What is the win? We get exception anyway

dwatl: Let's fail fast, maybe it will be more predictable but still not type safe :/

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dagguh: Records are allowed to have methods/behaviors. Who's encouraging anemic DTO classes?

dwatl: What's the advantage of doing this? Currently, AuroraCluster is just storing data and is immutable, easy to reason about.
Once you start adding more methods you have watch-out for internal state, mutability etc.

Not to mention you have to start adding annotations like `@JsonIgnore`, because you never know...

My thinking right now: _I don't need AuroraCluster in order to fetch server ids. It's already 100lines class, throwing more stuff in there won't make it more readable_

wyrzyk: > Not to mention you have to start adding annotations like `@JsonIgnore`. 

wat

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.3.0...master
 
+### Added
+- `RdsDns`, `AuroraCluster`, `AuroraEndpoint`, `AuroraReplicasDiscoverer`
```
wyrzyk, line 28: Do we need to expose all the classes in API?

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: I can move it to `AuroraCluster`, but shouldn't we move `fetchReplicasEndpoints` as well?

Anyway main pro is see: extracting `fetchReplicasServerIds` to a different place will help us to mock this dependency.

CHANGELOG.md
wyrzyk: Can we re-user common interfaces like `javax.sql.DataSource` instead of creating new ones?

I'd imagine we have an AuroraCluster and we can ask It for a main data source and replicaDataSources.

I don't understand the need for RdsDns, AuroraEndpoint and AuroraReplicasDiscoverer.

dwatl: Good question, it was tempted by this... All these classes are needed for com.atlassian.db.replica.api.aurora.AuroraEndpoints#instanceEndpoint
We can strip the abstraction/make it private. I will open another PR for this so we can compare.

dwatl: @wyrzyk are we good with keeping data classes package-private instead of making them public?
We could inline, but I'm worried it will make stuff less readable.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dagguh: Yes, we should move both fetches.
You can mock an entire `AuroraCluster`.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dagguh: Nice thinking outside the box! The function could read from a synthetic table. Mutating the table would be like re-mocking it for different test cases.

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,55 @@
...
+        this.connection = connection;
+    }
+
+    /**
+     * It provides list of replica endpoints based on reader endpoint and server ids
+     * @param readerEndpoint - Aurora reader endpoint
+     * @return list of replicas endpoints
+     * @throws SQLException
+     */
+    public List<AuroraEndpoint> fetchReplicasEndpoints(String readerEndpoint) throws SQLException {
```
dagguh, line 30: `AuroraReplicasDiscoverer` is public API so its public signatures are API too. This includes the `AuroraEndpoint` class.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoints.java
```diff
@@ -0,0 +1,20 @@
...
+import static com.atlassian.db.replica.internal.aurora.AuroraEndpoint.AuroraEndpointBuilder.anAuroraEndpoint;
+
+public class AuroraEndpoints {
+    private AuroraEndpoints() {
+    }
+
+    /**
+     * Transforms reader endpoint to instance endpoint
+     */
+    public static AuroraEndpoint instanceEndpoint(String readerEndpoint, String serverId) {
```
dagguh, line 13: So this procedural util function is equivalent to static `AuroraEndpoint.parse(String readerEndpoint, String serverId)`.

src/main/java/com/atlassian/db/replica/internal/aurora/RdsDns.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class RdsDns {
+    private static final Pattern DNS_PATTERN = Pattern.compile("([^.]+).([^:]*):?([0-9]+)?");
+    private final String region;
+    private final String domain;
+    private final Integer port;
+
+    public RdsDns(String region, String domain, Integer port) {
```
dagguh, line 13: Which ones are nullable? AFAIK all API so far was non-null. This means we need to start adding `@Nullable` and `@NonNull`.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraCluster.java
```diff
@@ -0,0 +1,100 @@
...
+
+
+    public static final class AuroraClusterBuilder {
+        private String clusterName;
+        private String clusterPrefix;
+
+        private AuroraClusterBuilder() {
+        }
+
+        public static AuroraClusterBuilder anAuroraCluster(String clusterName) {
```
dagguh, line 82: So `clusterPrefix` is nullable, but `clusterName` is not?

src/test/java/com/atlassian/db/replica/api/aurora/RdsDnsTest.java
```diff
@@ -0,0 +1,41 @@
...
+        assertThat(dns.getDomain()).isEqualTo("rds.amazonaws.com");
+        assertThat(dns.getPort()).isEqualTo(null);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+        "us-east-1.rds.amazonaws.com:3306",
+        "us-east-1.rds.amazonaws.com",
+    })
+    void shouldSerializerToOriginalValue(String input) {
```
dagguh, line 36: Nice one

src/main/java/com/atlassian/db/replica/internal/aurora/RdsDns.java
dwatl: Port can be nullable

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraCluster.java
dwatl: Correct

src/main/java/com/atlassian/db/replica/api/aurora/AuroraReplicasDiscoverer.java
dwatl: AuroraEndpoint -> String
or expose additional class 🤔 I don't mind returning string here, to be honest.

dwatl:
https://github.com/atlassian-labs/db-replica/pull/102

PR #89, opened 2021-06-17
Allow getting data from SuppliedCache without executing the supplier

wyrzyk:


src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -23,6 +23,11 @@
         return Optional.ofNullable(value);
     }
 
+    @Override
+    public Optional<T> get() {
+        return Optional.ofNullable(value);
```
dagguh, line 28: When is it ok to skip `maybeRefresh`?
The `Supplier` isn't executed if the lock is closed.

CHANGELOG.md
```diff
@@ -24,6 +24,18 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-1.3.0...master
 
+### Added
+- Allow getting cached value from `SuppliedCache` without executing a supplier
+
+### Changed
+
+- Make `PessimisticPropagationConsistency` class final
+- Make `ThrottledCache` class final
+- Make `NoCacheSuppliedCache` class final
```
dagguh, line 34: Breaking compatibility is ok 👌 
We just have to communicate it with a major bump. `db-replica:2.0.0` here we come!

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
wyrzyk: It's always ok to read the old value from the cache.

dagguh: Hmm, that's true for `SuppliedCache`, it never guaranteed any degree of freshness.

dagguh (approved):

PR #93, opened 2021-06-30
JCES-2122: Add timeout to ThrottledCache

wyrzyk:


tciesla-atlassian (approved):

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -29,12 +45,26 @@
     }
 
     private void maybeRefresh(Supplier<T> supplier) {
-        if(lock.tryLock()) {
-            try {
-                value = supplier.get();
-            } finally {
-                lock.unlock();
+        final boolean hasLock = lock.compareAndSet(null, Instant.now().plus(timeout));
```
dwatl, line 48: Would injecting a clock make it easier to test?

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -14,9 +16,23 @@
...
+        this.timeout = timeout;
+    }
+
+    /**
+     * It may be dangerous to use ThrottledCache without any timeout. A single supplier can block the cache refreshes forever.
+     * This constructor is available for compatibility only.
+     */
+    @Deprecated
+    public ThrottledCache() {
+        this.timeout = Duration.ofDays(365);
```
afaruga-atlassian, line 33: why not 1 Day?

wyrzyk: No reason.

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -29,12 +45,26 @@
...
-                lock.unlock();
+        final boolean hasLock = lock.compareAndSet(null, Instant.now().plus(timeout));
+        if (hasLock) {
+            loadValue(supplier);
+        } else {
+            final Instant instant = lock.get();
+            final boolean didLockExpire = instant != null && instant.isBefore(Instant.now());
+            if (didLockExpire) {
+                final boolean didLockTakeOver = lock.compareAndSet(instant, Instant.now().plus(timeout));
+                if (didLockTakeOver) {
```
afaruga-atlassian, line 56: no coverage in test for else  I mean: `didLockTakeOver = false`

wyrzyk: This case is hard to reproduce (we lose a race condition and another thread started to populate the cache). It would test if atomic reference keeps the contract. I'm not sure if there's enough value in that case to add a test case. 🤔

afaruga-atlassian: > It would test if atomic reference keeps the contract

afaruga-atlassian (approved):

PR #95, opened 2021-07-01
Fix throttled cache only one thread can update values

wyrzyk:


src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -57,36 +62,106 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
...
         public Optional<T> get() {
-            return Optional.ofNullable(value);
+            return Optional.ofNullable(lockValuePair.get().getValue());
         }
 
         private void maybeRefresh(Supplier<T> supplier) {
             final Instant deadline = clock.instant().plus(this.timeout);
-            final boolean hasLock = lock.compareAndSet(null, deadline);
+            final T currentValue = this.lockValuePair.get().getValue();
+            final LockValuePair<T> valueWithLock = new LockValuePair<>(
```
afaruga-atlassian, line 76: so you extended/refresh the deadline, this is like currentvalueWithRefreshedDeadline

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -57,36 +62,106 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
...
+                deadline,
+                currentValue
+            );
+            final LockValuePair<T> currentLockValuePair = this.lockValuePair.get();
+            final boolean hasLock = !currentLockValuePair.isLocked() && this.lockValuePair.compareAndSet(
```
afaruga-atlassian, line 81: here you compare the same value but for previous deadline and new one to get the lock.

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -57,36 +62,106 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
...
+        private void loadValue(Supplier<T> supplier, LockValuePair<T> currentLockValue) {
             try {
-                value = supplier.get();
-            } finally {
-                lock.compareAndSet(timeout, null);
+                final T newValue = supplier.get();
+                this.lockValuePair.compareAndSet(
+                    currentLockValue,
+                    new LockValuePair<>(
+                        null,
```
afaruga-atlassian, line 109: why null, not `now` ?

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -57,36 +62,106 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
...
+                        newValue
+                    )
+                );
+            } catch (Exception e) {
```
afaruga-atlassian, line 113: this will catch the exception that comes from supplier?

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
wyrzyk: null -> lock released
now() -> lock expired

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
wyrzyk: Yes

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -57,36 +62,106 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
...
+                    new LockValuePair<>(
+                        null,
+                        currentLockValue.getValue()
+                    )
+                );
+                throw e;
+            }
+        }
+
+        private static final class LockValuePair<T> {
```
afaruga-atlassian, line 125: so actually, what you did, you don't want to keep a lock on a clock, but on a value too. 
So you have a new object that help achieve this.

wyrzyk: Yes. I want the changes to be atomic.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -113,11 +114,58 @@ public void shouldNotRobbedThreadReleaseLock() throws InterruptedException {
...
+    @Test
+    public void shouldntRobbedThreadUpdateValue() throws InterruptedException {
+        ThrottledCache<Long> cache = new ThrottledCache<>(clock, ofMillis(1500));
+
+        final WaitingWork firstThreadWork = asyncPutWithSlowSupplier(cache, 1);
+        clock.tick();
+        clock.tick();
+        cache.get(() -> 2L);
+        firstThreadWork.finish();
+        assertThat(cache.get()).isEqualTo(Optional.of(2L));
```
afaruga-atlassian, line 126: good test, you want to enforce order of refreshed values.
You don't want that thread that comes first, but ended later, overridden the  value provided by second thread later, but finished earlier(without slow down).
👍

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -57,36 +62,106 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
...
+                if (oldLock.didLockExpire(clock)) {
+                    final LockValuePair<T> newLock = new LockValuePair<>(
+                        deadline,
+                        oldLock.getValue()
+                    );
+                    final boolean didLockTakeOver = lockValuePair
+                        .compareAndSet(oldLock, newLock);
                     if (didLockTakeOver) {
-                        loadValue(supplier, deadline);
+                        loadValue(supplier, newLock);
```
wyrzyk, line 97: There's one more thing that is tempting to implement. To `interrupt` the slow thread here (we are going to ignore the results anyway). The only thing that stops me is that I know it will be a database call and database calls are not interruptable, so there may be no benefit from the extra complexity.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -113,11 +114,58 @@ public void shouldNotRobbedThreadReleaseLock() throws InterruptedException {
...
+    public void shouldFailingSupplierReleaseTheLock() throws InterruptedException {
+        ThrottledCache<Long> cache = new ThrottledCache<>(clock, ofMillis(1500));
+
+        cache.get(() -> 1L);
+        try {
+            cache.get(() -> {
+                throw new RuntimeException();
+            });
+        } catch (Exception e) {
+            //ignore
```
afaruga-atlassian, line 139: for what `//ignore` refers to? exception or test scenario?

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -113,11 +114,58 @@ public void shouldNotRobbedThreadReleaseLock() throws InterruptedException {
...
+        }
+        final Optional<Long> valueAfterSupplierFailure = cache.get();
+        // this call expects no lock
```
afaruga-atlassian, line 142: I'm missing why it expects no lock.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -113,11 +114,58 @@ public void shouldNotRobbedThreadReleaseLock() throws InterruptedException {
...
+
+        assertThat(valueAfterSupplierFailure).isEqualTo(Optional.of(1L));
+    }
+
+    @Test
+    public void shouldTimeoutLockMultipleTimes() throws InterruptedException {
+        ThrottledCache<Long> cache = new ThrottledCache<>(clock, ofMillis(500));
+
+        final ArrayList<WaitingWork> waitingWorks = new ArrayList<>();
+        for (int i = 0; i < 32; i++) {
```
afaruga-atlassian, line 153: 2^5, why not 256? 🤣

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
wyrzyk: shouldFailingSupplierReleaseTheLock

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
afaruga-atlassian: > are not interruptable


well you never know, even if this / or sth else happens you can improve it along the way

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
```diff
@@ -113,11 +114,58 @@ public void shouldNotRobbedThreadReleaseLock() throws InterruptedException {
...
+        try {
+            cache.get(() -> {
+                throw new RuntimeException();
+            });
+        } catch (Exception e) {
+            //ignore
+        }
+        final Optional<Long> valueAfterSupplierFailure = cache.get();
+        // this call expects no lock
+        cache.get(() -> 2L);
```
afaruga-atlassian, line 143: I think this part is missing assertion. unless you check if it doesn't thrown an exception

afaruga-atlassian (dismissed):

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
wyrzyk: I refactored it.

src/test/java/com/atlassian/db/replica/api/ThrottledCacheTest.java
wyrzyk: 🤷

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
dwatl: DB calls can have timeout ;)

afaruga-atlassian (approved):

PR #96, opened 2021-07-08
Remove all assumptions from `DualConnection#close` and always close a…

wyrzyk:
…ll sub-connections


src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -154,6 +178,32 @@ public boolean hasWriteConnection() {
...
 
     public void close() throws SQLException {
+        if (compatibleWithPreviousVersion) {
+            close_old();
+        } else {
+            close_new();
+        }
+    }
+
+    public void close_new() throws SQLException {
```
dwatl, line 188: Don't we want to have new test cases for these?

wyrzyk: Test cases are green for both old and new, and _old (we don't have repro) will be removed just after the release.

dwatl (approved):
LGTM

PR #97, opened 2021-07-08
Close replica connection on `ReplicaConsistency#isConsistent` exception

wyrzyk:


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Dropping a requirement of a major version of a dependency is a new contract.
 - Only one thread can update `Throttled Cache` values
 - Remove all assumptions from `DualConnection#close` and always close all sub-connections
 - Fix connection leak
+- Close replica connection on `ReplicaConsistency#isConsistent` exception
```
afaruga-atlassian, line 31: This time, a couple of changes will be hidden behind `compatibleWithPreviousVersion`

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -262,7 +262,16 @@ private Connection prepareReadConnection(RouteDecisionBuilder decisionBuilder) t
             decisionBuilder.cause(writeConnection.getFirstCause().build());
             return writeConnection.get(decisionBuilder);
         }
-        if (consistency.isConsistent(() -> readConnection.get(decisionBuilder))) {
+        boolean isConsistent;
+        try {
+            isConsistent = consistency.isConsistent(() -> readConnection.get(decisionBuilder));
+        } catch (Exception e) {
+            if (!compatibleWithPreviousVersion) {
+                closeConnection(readConnection, decisionBuilder);
```
afaruga-atlassian, line 270: This is an interesting behaviour, it will cause to close and create replica connections more often.
For CBR it will occur more often, while for WAW? not much probably?
WDYT?

CHANGELOG.md
afaruga-atlassian: I wonder whether this change shouldnt be released as a separate. What side effects do you expect?

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: > This is an interesting behaviour, it will cause to close and create replica connections more often.
> For CBR it will occur more often, while for WAW? not much probably?
> WDYT?

It should never happen in our product as all our internal ReplicaConsistency implementations already handle exceptions. It could leak the connection if:
- ReplicaConsistency implementation doesn't catch it.
- and the usage of DualConnection is working only for happy path scenarios (for example connection is not closed in the finally block)

CHANGELOG.md
wyrzyk: I expect no side effects for us, but it can be helpful for other library consumers.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
afaruga-atlassian: If I understood correctly:

> and the usage of DualConnection is working only for happy path scenarios (for example connection is not closed in the finally block)

This is not happening as well in our product?

afaruga-atlassian (approved):

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: > This is not happening as well in our product?

It's happening, but not often.

PR #99, opened 2021-07-23
Add replica failure strategy

wyrzyk:
Currently, when DualConnection fails to get a replica connection,
the exception is propagated. In case of replica failure, there's
no way to fall back to the main database. The new SPI allows controlling failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-2.1.4...master
 
+### Added
+- Add `ReplicaFailureStrategy`
+
 ## [2.1.4] - 2021-07-08
 [2.1.4]: https://github.com/atlassian-labs/db-replica/compare/release-2.1.2...release-2.1.4
 
--- a/src/main/java/com/atlassian/db/replica/api/reason/Reason.java
+++ b/src/main/java/com/atlassian/db/replica/api/reason/Reason.java
@@ -32,6 +32,8 @@ public final class Reason {
             new ReasonBuilder("HIGH_TRANSACTION_ISOLATION_LEVEL").isRunOnMain(true).isWrite(false).build();
     public static final Reason RO_API_CALL =
             new ReasonBuilder("RO_API_CALL").isRunOnMain(false).isWrite(false).build();
+    public static final Reason REPLICA_GET_FAILURE =
+        new ReasonBuilder("REPLICA_GET_FAILURE").isRunOnMain(true).isWrite(false).build();
 
     public String getName() {
         return name;
@@ -75,7 +77,7 @@ public final class Reason {
         ReasonBuilder(final String name) {
             this.name = name;
         }
-        
+
         ReasonBuilder isRunOnMain(boolean isRunOnMain) {
             this.isRunOnMain = isRunOnMain;
             return this;
--- a/src/main/java/com/atlassian/db/replica/api/strategy/FallBackToMain.java
+++ b/src/main/java/com/atlassian/db/replica/api/strategy/FallBackToMain.java
@@ -1,0 +1,18 @@
+package com.atlassian.db.replica.api.strategy;
+
+import com.atlassian.db.replica.api.SqlCall;
+import com.atlassian.db.replica.spi.ReplicaFailureStrategy;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/**
+ * Usees main connection after replica failure.
+ */
+public class FallBackToMain implements ReplicaFailureStrategy {
+
+    @Override
+    public Connection onFailure(SQLException exception, SqlCall<Connection> mainConnection) throws SQLException {
+        return mainConnection.call();
+    }
+}
--- a/src/main/java/com/atlassian/db/replica/api/strategy/PropagateReplicaFailure.java
+++ b/src/main/java/com/atlassian/db/replica/api/strategy/PropagateReplicaFailure.java
@@ -1,0 +1,17 @@
+package com.atlassian.db.replica.api.strategy;
+
+import com.atlassian.db.replica.api.SqlCall;
+import com.atlassian.db.replica.spi.ReplicaFailureStrategy;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/**
+ * Propagates replica failure.
+ */
+public class PropagateReplicaFailure implements ReplicaFailureStrategy {
+    @Override
+    public Connection onFailure(SQLException exception, SqlCall<Connection> mainConnection) throws SQLException {
+        throw exception;
+    }
+}
--- a/src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
+++ b/src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
@@ -8,6 +8,7 @@ import com.atlassian.db.replica.internal.SqlRunnable;
 import com.atlassian.db.replica.internal.Warnings;
 import com.atlassian.db.replica.spi.ConnectionProvider;
 import com.atlassian.db.replica.spi.ReplicaConsistency;
+import com.atlassian.db.replica.spi.ReplicaFailureStrategy;
 
 import java.sql.Connection;
 import java.sql.SQLException;
@@ -17,6 +18,7 @@ import java.util.concurrent.Executor;
 
 import static com.atlassian.db.replica.api.reason.Reason.HIGH_TRANSACTION_ISOLATION_LEVEL;
 import static com.atlassian.db.replica.api.reason.Reason.MAIN_CONNECTION_REUSE;
+import static com.atlassian.db.replica.api.reason.Reason.REPLICA_GET_FAILURE;
 import static com.atlassian.db.replica.api.reason.Reason.REPLICA_INCONSISTENT;
 import static com.atlassian.db.replica.api.reason.Reason.RO_API_CALL;
 import static com.atlassian.db.replica.api.reason.Reason.RW_API_CALL;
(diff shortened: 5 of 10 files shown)


afaruga-atlassian:
@wyrzyk Why you closed it? Any other reason than none review? If so I'll review it soon. If other please share the reasons to close it.

wyrzyk:
@afaruga-atlassian It's not complete yet, and I didn't know how to convert it to draft. I re-opened it, and I'll add a task for the missing part.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -53,9 +54,16 @@ public ConnectionState(
...
                 if (connectionProvider.isReplicaAvailable()) {
-                    final Connection replicaConnection = connectionProvider.getReplicaConnection();
-                    parameters.initialize(replicaConnection);
-                    return replicaConnection;
+                    try {
+                        final Connection replicaConnection = connectionProvider.getReplicaConnection();
+                        parameters.initialize(replicaConnection);
+                        return replicaConnection;
+                    } catch (SQLException e) {
+                        return replicaFailureStrategy.onFailure(
```
wyrzyk, line 62: After the change readConnection was pointing to the main database connection only when the replica was disabled. The change may affect:
- connection state.
- isConsistent checks.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-2.1.4...master
 
+### Added
+- Add `ReplicaFailureStrategy`
```
afaruga-atlassian, line 28: It would be good to explain what does it mean FailureStrategy ? 
is it circuit breaker? fallback ? what kind of failure? any failure?

src/main/java/com/atlassian/db/replica/api/strategy/FallBackToMain.java
```diff
@@ -0,0 +1,18 @@
...
+import java.sql.SQLException;
+
+/**
+ * Usees main connection after replica failure.
+ */
+public class FallBackToMain implements ReplicaFailureStrategy {
+
+    @Override
+    public Connection onFailure(SQLException exception, SqlCall<Connection> mainConnection) throws SQLException {
+        return mainConnection.call();
```
afaruga-atlassian, line 16: How to add any observability here?

src/main/java/com/atlassian/db/replica/api/strategy/PropagateReplicaFailure.java
```diff
@@ -0,0 +1,17 @@
...
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/**
+ * Propagates replica failure.
+ */
+public class PropagateReplicaFailure implements ReplicaFailureStrategy {
+    @Override
+    public Connection onFailure(SQLException exception, SqlCall<Connection> mainConnection) throws SQLException {
+        throw exception;
```
afaruga-atlassian, line 15: what if exception would be other type than `SQLException`

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
afaruga-atlassian, line 62: have you thought about more sophisticated mechanism aka CirucitBreaker configuration:
How often it will be run ? on every failed call ? fallback only after 5 failed calls, happened on x concurrent threads, and lasting x minutes? 
Should it retry  ?
I've just listed my ideas, doesn't insist to implement all of them, more want to understood your approach for them.

src/main/java/com/atlassian/db/replica/spi/ReplicaFailureStrategy.java
```diff
@@ -0,0 +1,14 @@
...
+import com.atlassian.db.replica.api.SqlCall;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public interface ReplicaFailureStrategy {
+    /**
+     * Controls replica failures.
+     */
+    Connection onFailure(SQLException exception, SqlCall<Connection> mainConnection) throws SQLException;
```
afaruga-atlassian, line 13: looking only from the perspective of  `ReplicaFailureStrategy` API it looks that `replicaConnection` is skipped.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -191,6 +192,51 @@ public void shouldSwitchToMainIfNotConsistent() throws SQLException {
...
+            public Connection getReplicaConnection() throws SQLException {
+                throw new SQLException("Can't connect to the replica");
+            }
+        };
+        final DatabaseCall databaseCall = mock(DatabaseCall.class);
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).databaseCall(databaseCall)
+            .replicaFailureStrategy(new FallBackToMain())
```
afaruga-atlassian, line 227: formatting

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
```diff
@@ -32,6 +32,8 @@ private Reason(final String name, boolean isRunOnMain, boolean isWrite) {
             new ReasonBuilder("HIGH_TRANSACTION_ISOLATION_LEVEL").isRunOnMain(true).isWrite(false).build();
     public static final Reason RO_API_CALL =
             new ReasonBuilder("RO_API_CALL").isRunOnMain(false).isWrite(false).build();
+    public static final Reason REPLICA_GET_FAILURE =
```
afaruga-atlassian, line 35: Why `REPLICA_GET_FAILURE` not `REPLICA_FAILURE`

afaruga-atlassian (changes requested):
- [ ] Add IT test which would disable docker instance under the cover

src/main/java/com/atlassian/db/replica/api/strategy/FallBackToMain.java
afaruga-atlassian: Ok, the customer should decorate `ReplicaFailureStrategy` too add obsverability

afaruga-atlassian:
> @afaruga-atlassian It's not complete yet, and I didn't know how to convert it to draft. I re-opened it, and I'll add a task for the missing part.

👍 Sure
re converting to draft:
https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/
Screenshot 2021-07-27 at 14 28 25

CHANGELOG.md
wyrzyk: Replica failure = DualConnection fails to get a replica connection.

src/main/java/com/atlassian/db/replica/api/strategy/PropagateReplicaFailure.java
wyrzyk: `SQLException` is part of the `Connection` API and `ConnectionProvider` SPI.  A library customer provides ConnectionProvider implementation. It's up to the customer to ensure the contract.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: > have you thought about more sophisticated mechanism

Yes. That's why it's an SPI.

src/main/java/com/atlassian/db/replica/spi/ReplicaFailureStrategy.java
wyrzyk: > looking only from the perspective of ReplicaFailureStrategy API it looks that replicaConnection is skipped.

Replica connection supplier failed and thrown an exception. We have an exception and a supplier for the main connection. What scenario would require a replicaConnection supplier?

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
wyrzyk: I think it better describes the reason. We failed to get a replica connection. Maybe replica failed, maybe the connection pool is saturated.

wyrzyk:
>  Add IT test which would disable docker instance under the cover

Why enforcing an IT test when the behaviour can be covered with a unit test?

CHANGELOG.md
afaruga-atlassian: Perhaps it's worth to add this to docs, there.

afaruga-atlassian: And thanks for explaining, this is different than  db queries failing while using a replica connection , that would be more a CircuitBreaker  IIUC

src/main/java/com/atlassian/db/replica/spi/ReplicaFailureStrategy.java
afaruga-atlassian: Retrying to use replicaConnection,

src/main/java/com/atlassian/db/replica/api/reason/Reason.java
afaruga-atlassian: `REPLICA_GET_CONNECTION_FAILURE` 

I didn't known it is about connection, I thought it's about replica _any_ failure

afaruga-atlassian:
> > Add IT test which would disable docker instance under the cover
> 
> Why enforcing an IT test when the behaviour can be covered with a unit test?

That depends, if unit tests go through all layers (if it had too) .
In case when unit test mock some mechanism, which will be different on PRODUCTION there is a chance that it will be false-positive scenario which want test the real thing.
That's why having 1 IT tests which actually use all the layers in between can be beneficial.

Not sure if this apply here, that's why I left the comment.

wyrzyk:
> Not sure if this apply here, that's why I left the comment.

The unit test ensures a contract between `ConnectionProvider` and `ReplicaFailureStrategy`. Both are provided by a library customer. We can't do integration tests here (even if we do, it would integration test our internal mock implementations).

afaruga-atlassian:
> it would integration test our internal mock implementations

Do you mean that the proper place for IT tests is customer code?

afaruga-atlassian:
Hey @wyrzyk I've just realized that I didn't  shared previously my answer. I don't consider 
> Add IT test which would disable docker instance under the cover

as a blocker, but once you request changes, then there is no way to de-request it. The only way is to approve.
I didn't approve as there are 2 tasks added by you. Once they're done I'm happy to approve.

PR #102, opened 2021-08-06
[DRAFT] DB replica discovery.

wyrzyk:
I started from API and API tests. Then copy pasted part of Dawid's PR
(https://github.com/atlassian-labs/db-replica/pull/88/files#r650820456).

TODO:
- Improve API.
- Add more API level tests.
- Copy paste rest of the Dawid's work.
- Refactor.


src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
```diff
@@ -0,0 +1,10 @@
+package com.atlassian.db.replica.spi;
+
+import com.atlassian.db.replica.api.Database;
+
+import java.sql.SQLException;
+import java.util.stream.Stream;
+
+public interface DatabaseCluster {
```
wyrzyk, line 8: This is the most important part of the pull-request (API). @dwatl Do you think it will expose enough information to build replica-discovery?

dwatl: How do we plan to implement Database? We need cluster URL and replica id

wyrzyk: Cluster URL seems to be a constant for the cluster, and replica id can be fetched from the database. Am I wrong? Can you elaborate?

src/main/java/com/atlassian/db/replica/api/AuroraCluster.java
```diff
@@ -0,0 +1,33 @@
...
+
+        return discoverer.fetchReplicasServerIds().stream().map(it -> new Database() {
+            @Override
+            public String getUuid() {
+                return it;
+            }
+
+            @Override
+            public Connection getConnection() {
+                return null;
```
dwatl, line 28: 😨 😄

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
```diff
@@ -0,0 +1,10 @@
...
+    Stream<Database> getReplicas() throws SQLException;
```
dwatl, line 9: I think we'd rather have a collection here

src/main/java/com/atlassian/db/replica/api/Database.java
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.db.replica.api;
+
+import java.sql.Connection;
+
+public interface Database {
+    String getUuid();
```
dwatl, line 6: getUuid -> getId? Otherwise it should return UUID not String

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
wyrzyk: Why?

src/main/java/com/atlassian/db/replica/api/AuroraCluster.java
```diff
@@ -0,0 +1,33 @@
...
+    public Stream<Database> getReplicas() throws SQLException {
+
+        return discoverer.fetchReplicasServerIds().stream().map(it -> new Database() {
+            @Override
+            public String getUuid() {
+                return it;
+            }
+
+            @Override
+            public Connection getConnection() {
```
wyrzyk, line 27: I think it should return a supplier. It depends on the client, but this way we could cache connections to known replicas and fetch only the newly discovered replicas connections.

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
dwatl: Hmm, correct!

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
dwatl: Streams can be potentially infinite, you cannot use them more than once

src/main/java/com/atlassian/db/replica/api/Database.java
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.db.replica.api;
+
+import java.sql.Connection;
+
+public interface Database {
```
afaruga-atlassian, line 5: How about adding method: `getRole() `
`Reader`, `Writer`

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraCluster.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private AuroraCluster(String clusterName, String clusterPrefix) {
+        this.clusterName = clusterName;
+        this.clusterPrefix = clusterPrefix;
+    }
+
+    public static AuroraCluster parse(String cluster) {
+        Objects.requireNonNull(cluster);
+
+        String[] chunks = splitByLastOccurence("-", cluster);
```
afaruga-atlassian, line 19: Whether this string parsing is Aurora specific? or would it also work on Postgres?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoint.java
```diff
@@ -0,0 +1,88 @@
+package com.atlassian.db.replica.internal.aurora;
+
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class AuroraEndpoint {
+    private static final Pattern ENDPOINT_PATTERN = Pattern.compile("([^.]+).([^.]+).(.*)");
```
afaruga-atlassian, line 8: Does it comming from AWS naming endpoints? or from Atlassian-internal-service?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Asks Aurora about the list of server ids
+     * @return server ids
+     * @throws SQLException
+     */
+    public List<String> fetchReplicasServerIds() throws SQLException {
+        LinkedList<String> ids = new LinkedList<>();
+        try (ResultSet rs =
+                 connection.prepareStatement(
+                     "SELECT server_id FROM aurora_global_db_instance_status() WHERE session_id != 'MASTER_SESSION_ID'")
```
afaruga-atlassian, line 44: I wonder whether aurora specific functions have to be used, or postgress would be also compatible ?
If so I would be for postgress, why ? aaurora is postgress compatible and it can be easier test on postgres-docker

src/test/java/com/atlassian/db/replica/api/AuroraClusterMock.java
```diff
@@ -0,0 +1,105 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.sql.Types;
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class AuroraClusterMock {
+    private final Connection connection = DriverManager.getConnection("jdbc:h2:mem:testdb", "sa", "");
```
afaruga-atlassian, line 18: why not `IT tests` that run postgress on docker with replica ?

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
afaruga-atlassian: +1 for Collection.
Why Streams?

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
afaruga-atlassian, line 8: DatabaseCluster should contain all databases, MAIN + REPLICAS or Writer with Readers

afaruga-atlassian: Other than that, databaseEngine - as it might be Aurora, might be not.

src/main/java/com/atlassian/db/replica/api/AuroraCluster.java
afaruga-atlassian: With supplier  the customer can define  Eager/lazy initialization

src/main/java/com/atlassian/db/replica/api/Database.java
wyrzyk: Please explain the use case.

src/test/java/com/atlassian/db/replica/api/AuroraClusterMock.java
wyrzyk: Emulating Aurora API is easier/faster without docker.

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
wyrzyk: > DatabaseCluster should contain all databases, MAIN + REPLICAS or Writer with Readers

No use for the main database yet.

> Other than that, getDatabaseEngine - as it might be Aurora, might be not.

What use case do you mean? The implementation defines the database/database engine.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicasDiscoverer.java
afaruga-atlassian:
```    
assertThat(executeQueryAndReturnWithMetaData(postgressProvider.getPureMainConnection(), "SELECT pg_is_in_recovery()")).isEqualTo("f pg_is_in_recovery");
assertThat(executeQueryAndReturnWithMetaData(postgressProvider.getPureReplicaConnection(), "SELECT pg_is_in_recovery()")).isEqualTo("t pg_is_in_recovery");
```

`select pg_backend_pid();`

https://www.postgresql.org/message-id/BANLkTimwRLFOPriKoUyG6DESTTLBOa_N3Q@mail.gmail.com

https://www.postgresql.org/docs/9.1/functions-info.html

src/test/java/com/atlassian/db/replica/api/AuroraClusterMock.java
afaruga-atlassian: >Emulating Aurora API is easier/faster without docker.

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
afaruga-atlassian: Ok, to be fair, the first thing I did I looked into TCS into customer code and looked what it know about DatabaseCluster.

you're right with focusing on API usecases first.

src/main/java/com/atlassian/db/replica/spi/DatabaseCluster.java
wyrzyk: Streams or Iterables give more flexibility to implementations. I thought about future implementations that could create connections lazily, but I think `Supplier<Connection>` will work even better. For aurora we know we can get data with a single call, so no benefit from lazily initialized Database stream. Will it be true for other database clusters? Will anyone need it? I don't know.

Usually, it's safer to start with a stream/iterable API when the collection can be large. This way you can fix performance problems without breaking the API. I think this time we can use Collection and see how it evolves.

wyrzyk: > can be potentially infinite,

The consumer controls the flow, so I see no problem with that.


>  you cannot use them more than once

It's something everyone should be aware of. In case you need to use it multiple times, you store it in a collection.

dwatl: I'm fine with Collection 👌

PR #104, opened 2021-08-13
Add cluster support

wyrzyk:


src/main/java/com/atlassian/db/replica/spi/ClusterConsistency.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    default void preCommit(Connection main) throws SQLException {
+    }
+
+    /**
+     * Judges if {@code replica} is ready to be queried.
+     *
+     * @return true if {@code replica} is consistent with main
+     */
+    boolean isConsistent(Collection<Database> replicas) throws SQLException;
```
dwatl, line 35: This will be hiding things like "one replica constantly lagging" behind a single boolean value. Did you consider how it could affect debuggability?

wyrzyk: What do you mean by hiding? It's SPI.

dwatl (approved):

PR #105, opened 2021-08-13
Get rid of hardcoded jdbc url

dwatl:


src/main/java/com/atlassian/db/replica/internal/aurora/AuroraJdbcUrl.java
```diff
@@ -0,0 +1,25 @@
+package com.atlassian.db.replica.internal.aurora;
+
+public class AuroraJdbcUrl {
```
wyrzyk, line 3: It's internal but it's a parameter for the API constructor (AuroraCluster). Please make it API or fix the constructor to use API only classes.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraJdbcUrl.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final String prefix = "jdbc:postgresql://";
+    private final AuroraEndpoint endpoint;
+    private final String databaseName;
+
+    public AuroraJdbcUrl(AuroraEndpoint endpoint, String databaseName) {
```
wyrzyk, line 8: The same as above. Is `AuroraEndpoint` API or not?

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
```diff
@@ -61,22 +65,22 @@ public void shouldUtilizeReplicaForReadQueriesForSynchronisedWrites() throws SQL
     private SqlCall<Connection> initializeConnectionPool(final DatabaseCall decisionLog) throws SQLException {
         final ReplicaNode replicaNode = new ReplicaNode();
         final ConnectionProvider connectionProvider = new AuroraConnectionProvider(
```
wyrzyk, line 67: Should `AuroraConnectionProvider` accept Strings?

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
```diff
@@ -31,8 +33,10 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class AuroraClusterTest {
-    final String writerEndpoint = "jdbc:postgresql://database-1.cluster-crmnlihjxqlm.eu-central-1.rds.amazonaws.com:5432/newdb";
-    final String readerEndpoint = "jdbc:postgresql://database-1.cluster-ro-crmnlihjxqlm.eu-central-1.rds.amazonaws.com:5432/newdb";
+    final AuroraJdbcUrl writerUrl = new AuroraJdbcUrl(
+        AuroraEndpoint.parse("database-1.cluster-crmnlihjxqlm.eu-central-1.rds.amazonaws.com:5432"), "newdb");
```
wyrzyk, line 37: Can we avoid static methods in API?

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
```diff
@@ -6,6 +6,8 @@
 import com.atlassian.db.replica.api.aurora.ReplicaNode;
 import com.atlassian.db.replica.api.reason.Reason;
 import com.atlassian.db.replica.api.reason.RouteDecision;
+import com.atlassian.db.replica.internal.aurora.AuroraEndpoint;
```
wyrzyk, line 9: This is an example of lib usage. It should avoid calling internals.

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
```diff
@@ -31,8 +33,10 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class AuroraClusterTest {
-    final String writerEndpoint = "jdbc:postgresql://database-1.cluster-crmnlihjxqlm.eu-central-1.rds.amazonaws.com:5432/newdb";
-    final String readerEndpoint = "jdbc:postgresql://database-1.cluster-ro-crmnlihjxqlm.eu-central-1.rds.amazonaws.com:5432/newdb";
+    final AuroraJdbcUrl writerUrl = new AuroraJdbcUrl(
```
wyrzyk, line 36: Can it be simplified? (for example, created in one step if we never need the indirect AuroraEndpoint object)

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
dwatl: What's wrong with static methods in API?

wyrzyk: Nothing 😉.

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
dwatl: I gotta rethink ConnectionProvider, I wonder if it should be more multi-replica friendly

PR #106, opened 2021-08-16
Multi replica support

wyrzyk:


src/main/java/com/atlassian/db/replica/spi/ClusterConsistency.java
```diff
@@ -0,0 +1,36 @@
+package com.atlassian.db.replica.spi;
+
+import com.atlassian.db.replica.api.Database;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.Collection;
+
+public interface ClusterConsistency {
```
wyrzyk, line 9: Something is wrong with the API. Even if we use a cluster, it's enough to use ReplicaConsistency for CBR (check before read). `ClusterConsistency` seems to be a good fit only for cluster + WAW (Synchronous writes). It seems it's not a good abstraction to replace ReplicaConsistency. Maybe it's worth keeping ReplicaConsistency as a consistency API and see how it evolves in the future.

dwatl: @wyrzyk Why do we need ClusterConsistency? If you combine replica discoverer and ReplicaConsistency you have the ability to check consistency for each replica / filter only consistent replicas etc. Do we really need contracts for cluster?

wyrzyk: >  If you combine replica discoverer and ReplicaConsistency 

I don't want to introduce discoverer API. I'm still working on the PR to simplify it more. I think the new API will be only a new method in ConnectionProvider (generalized getReplicaConnection to getReplicas) + ClusterConsistency. ClusterConsistency may not be a good name 🤔. I like the change of return type from boolean to Optional<Database>. It seems to enable more use-cases:
- CBR can either use AWS LB to resolve the connection or use its own LB and provide one of the already consistent replicas.
- Same for WAW, we don't need to wait for all replicas. We just need to ensure ConsistencyProvider write and getConsistent are consistent (if it waits only for 3 replica nodes, it can LB only between the 3 most up to date replicas).

dwatl: > if it waits only for 3 replica nodes, it can LB only between the 3 most up to date replicas
Does it make sense when it comes to distributed systems? I think about situations where you have multiple-nodes. Local guarding is not enough if you want to keep consistency.

wyrzyk: > Local guarding is not enough if you want to keep consistency

Is it local when with replica discovery? All nodes see the same state of cluster topology.

dwatl: > We just need to ensure ConsistencyProvider write and getConsistent are consistent
Hmm, this would make WAW to check before reading. Which makes it CBR, right?

wyrzyk: It would be WAW but wait for some replicas instead of wait for all.

wyrzyk: Anyway, the API is still broken and will evolve.

dwatl: > It would be WAW but wait for some replicas instead of wait for all.
How would we guarantee that we only read from consistent replicas? It would require some checking before reading, right?

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -0,0 +1,44 @@
+package com.atlassian.db.replica.api;
+
+public final class AuroraConnectionDetails {
```
wyrzyk, line 3: There's nothing Aurora specific here. @tciesla-atlassian WDYT about renaming it to ConnectionDetails?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoints.java
```diff
@@ -0,0 +1,19 @@
...
+    private AuroraEndpoints() {
+    }
+
+    /**
+     * Transforms reader endpoint to instance endpoint
+     */
+    public static AuroraEndpoint instanceEndpoint(AuroraEndpoint readerEndpoint, String serverId) {
+        return anAuroraEndpoint(readerEndpoint)
+            .serverId(serverId)
+            .cluster(anAuroraCluster(readerEndpoint.getCluster().getClusterName()).clusterPrefix(null).build())
```
wyrzyk, line 16: What's the cluster prefix. What does it mean `clusterPrefix(null)`?

wyrzyk:
Update Changelog:

- [ ] AuroraMultiReplciaConsistency
- [ ] AuroraConnectionDetails

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/api/AuroraSequence.java
```diff
@@ -0,0 +1,49 @@
+package com.atlassian.db.replica.it.example.aurora.replica.api;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+
+public final class AuroraSequence {
```
wyrzyk, line 7: - [ ] Remove comments

src/main/java/com/atlassian/db/replica/api/Database.java
```diff
@@ -0,0 +1,10 @@
+package com.atlassian.db.replica.api;
+
+import java.sql.Connection;
+import java.util.function.Supplier;
+
+public interface Database {
```
wyrzyk, line 6: - [ ] Move to internal

src/main/java/com/atlassian/db/replica/internal/DecisionAwareReference.java
```diff
@@ -20,6 +20,11 @@ public void reset() {
         firstCause.set(null);
     }
 
+    public void set(T reference) {
```
wyrzyk, line 23: - [ ] Remove it

src/main/java/com/atlassian/db/replica/internal/LazyReference.java
```diff
@@ -33,6 +33,12 @@ public T get() {
         return reference;
     }
 
+    protected void set(T reference) {
```
wyrzyk, line 36: - [ ] Remove it

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/spi/ExceptionSpecificRunnable.java
```diff
@@ -0,0 +1,12 @@
+package com.atlassian.db.replica.it.example.aurora.replica.spi;
+
+/**
+ * An interface that is similar to a {@link Runnable} except that the {@link #run()} method can throw a specific checked
+ * exception.
+ * @param <E> the type of the exception that may be thrown
+ */
+public interface ExceptionSpecificRunnable<E extends Exception> {
```
wyrzyk, line 8: - [ ] Remove?

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/spi/Latency.java
```diff
@@ -0,0 +1,5 @@
+package com.atlassian.db.replica.it.example.aurora.replica.spi;
+
+public interface Latency {
```
wyrzyk, line 3: Remove?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoint.java
```diff
@@ -0,0 +1,92 @@
...
+        this.serverId = serverId;
+        this.cluster = cluster;
+        this.dns = dns;
+    }
+
+    public static AuroraEndpoint parse(String readerEndpoint) {
+        Objects.requireNonNull(readerEndpoint);
+
+        Matcher matcher = ENDPOINT_PATTERN.matcher(readerEndpoint);
+        matcher.matches();
```
wyrzyk, line 27: - [ ] verify matches() results

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraEndpoint.java
```diff
@@ -0,0 +1,92 @@
...
+        return new AuroraEndpoint(
+            matcher.group(1),
+            AuroraCluster.parse(matcher.group(2)),
+            RdsDns.parse(matcher.group(3))
+        );
+    }
+
+    public String getServerId() {
```
wyrzyk, line 35: - [ ] getters are never used

src/main/java/com/atlassian/db/replica/internal/aurora/ReplicaNode.java
```diff
@@ -0,0 +1,28 @@
...
+
+
+public class ReplicaNode {
+    private static final String AURORA_REPLICA_ID = "replicaId";
+
+    public Connection mark(final Connection connection, final String repliacId) {
+        try {
+            connection.getClientInfo().setProperty(AURORA_REPLICA_ID, repliacId);
+        } catch (SQLException throwables) {
+            //TODO:            log.withoutCustomerData().error("Failed to label a connection", e);
```
wyrzyk: - [ ] throw exception
- [ ] remove comments

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/api/SequenceReplicaConsistency.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+import static java.time.Duration.ofSeconds;
+
+//TODO promote it to the API
+//TODO add logs
+@SuppressWarnings({"SqlNoDataSourceInspection", "SqlResolve"})
+public class SequenceReplicaConsistency implements ReplicaConsistency {
```
wyrzyk, line 22: - [ ] cleanup the class

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/api/SynchronousWriteConsistency.java
```diff
@@ -0,0 +1,138 @@
...
+import static java.lang.Math.min;
+
+/**
+ * Waits until writes propagate to replica.
+ * It doesn't depend on a cross-JVM cache.
+ */
+
+// TODO add also DeferredSynchronousWriteConsistency implementation
+// TODO multi replica support
+public class SynchronousWriteConsistency implements ReplicaConsistency {
```
wyrzyk, line 23: - [ ] cleanup

tciesla-atlassian (approved):

PR #107, opened 2021-08-16
Simplify cluster api

wyrzyk:


src/main/java/com/atlassian/db/replica/internal/ConsistencyAdapter.java
```diff
@@ -26,16 +27,13 @@ public void preCommit(Connection main) {
     }
 
     @Override
-    public boolean isConsistent(Collection<Database> replicas) throws SQLException {
+    public Optional<Database> getConsistent(Collection<Database> replicas) throws SQLException {
```
dwatl, line 30: WDYT Optional -> Collection?

PR #110, opened 2021-08-18
Remove ClusterConsistency

wyrzyk:
Remove ReplicaNodeAwareConnectionProvider
Remove AuroraCluster to AuroraClusterDiscovery
Move AuroraClusterDiscovery to internal


src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
```diff
@@ -63,21 +61,18 @@ void shouldUtilizeReplicaForReadQueriesForSynchronisedWrites() throws SQLExcepti
...
-        final DatabaseCluster cluster = new AuroraCluster(
-            connectionProvider::getMainConnection,
-            readerEndpoint,
-            databaseName
-        );
         final ReplicaConsistency replicaConsistency = new ConsistencyFactory(
             connectionProvider::getMainConnection,
             connectionProvider,
-            cluster
+            readerEndpoint,   // TODO: I added the params here, but I'm sure databaseName can be discovered, not sure about readerEndpoint.
```
dwatl, line 67: > I added the params here, but I'm sure databaseName can be discovered, not sure about readerEndpoint.

It would require an additional call to db. I would lean more towards a solution where we don't need to pass connection for discovery. I think it serves as optimization, but once we implement an internal pool it will become obsolete.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -224,32 +215,18 @@ private Connection prepareReadConnection(RouteDecisionBuilder decisionBuilder) t
...
-                    }
-                });
-            consistentReplica = consistency.getConsistent(replicas);
+            isConsistent = consistency.isConsistent(() -> readConnection.get(decisionBuilder));
         } catch (Exception e) {
             closeConnection(readConnection, decisionBuilder);
             throw e;
         }
-        if (consistentReplica.isPresent()) {
+        if (isConsistent) {
```
dwatl, line 225: This is a bit long. Perhaps we could extract the if to a private method?

dwatl (approved):
🚀

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
wyrzyk: > It would require an additional call to db.

It's just one call. It can be cached as it doesn't change.

> I would lean more towards a solution where we don't need to pass connection for discovery. 

Can you elaborate? How can we discover cluster topology without a connection?

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: It' just reverts to the previous state. It can be refactored on a separate PR.

PR #112, opened 2021-08-27
Simple refactor of AuroraClusterDiscovery with few improvements.

tciesla-atlassian:


src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
```diff
@@ -61,7 +64,12 @@ void shouldUtilizeReplicaForReadQueriesForSynchronisedWrites() throws SQLExcepti
             readerJdbcUrl,
             writerJdbcUrl
         );
-        final ReplicaConsistency replicaConsistency = new ConsistencyFactory(connectionProvider).create();
+
+        Properties properties = new Properties();
```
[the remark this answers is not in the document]
tciesla-atlassian: Now it should work.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -37,7 +37,7 @@ public AuroraReplicasDiscoverer(AuroraJdbcUrl readerUrl) {
     }
 
     private List<String> fetchReplicasServerIds(Connection connection) throws SQLException {
-        LinkedList<String> ids = new LinkedList<>();
+        List<String> ids = new LinkedList<>();
```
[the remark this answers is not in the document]
tciesla-atlassian: Putting author shoes i would guess that these are the reasons:
-  we don't know up front how many results will be in ResultSet, so it's not possible to initialise ArrayList with correct size
- iterator in calling method is used to walk throughout all collection so random access is not necessary
IMO type of collection doesn't matter here, because we will have up to 15 replicas (scale is pretty low)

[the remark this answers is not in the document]
wyrzyk: > we don't know up front how many results will be in ResultSet, so it's not possible to initialise ArrayList with correct size

We know it's up to 15 and the default for ArrayList is 16 :)


> IMO type of collection doesn't matter here

You're right. In most cases, ArrayList is a better choice, so I was just curious why to use LinkedList here.

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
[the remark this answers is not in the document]
tciesla-atlassian: It's not a target solution - we want to extract connection creation out of library scope, so this class will disappear. Properties instance give us more flexibility (it doesn't force us to make any new changes with new property), but it comes with significant disadvantages as you listed above. We should promote readability and make code easier to use so IMO it's good idea to create our own type.

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -0,0 +1,23 @@
+package com.atlassian.db.replica.api;
+
+import java.util.Properties;
+
+public class AuroraConnectionDetails {
```
wyrzyk, line 5: Make it final.

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public static AuroraConnectionDetails of(String username, String password) {
+        Properties properties = new Properties();
+        properties.setProperty("user", username);
+        properties.setProperty("password", password);
+        return new AuroraConnectionDetails(properties);
+    }
+
+    public Properties getProperties() {
```
wyrzyk, line 19: It's public in API package = its API

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -1,30 +1,28 @@
...
+package com.atlassian.db.replica.api;
 
 import com.atlassian.db.replica.internal.aurora.AuroraClusterDiscovery;
 import com.atlassian.db.replica.spi.ReplicaConsistency;
 
 import java.sql.Connection;
 import java.util.function.Supplier;
 
-//TODO: promote to API?
 public class AuroraMultiReplicaConsistency implements ReplicaConsistency {
```
wyrzyk, line 9: final

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -1,30 +1,28 @@
...
     private final ReplicaConsistency replicaConsistency;
     private final AuroraClusterDiscovery cluster;
 
     public AuroraMultiReplicaConsistency(
```
wyrzyk, line 13: It will be harder to evolve constructors than builders in the future.

src/main/java/com/atlassian/db/replica/internal/aurora/ReplicaNode.java
```diff
@@ -7,12 +7,11 @@
...
-    public Connection mark(final Connection connection, final String repliacId) {
+    public Connection mark(final Connection connection, final String replicaId) {
         try {
-            connection.getClientInfo().setProperty(AURORA_REPLICA_ID, repliacId);
-        } catch (SQLException throwables) {
-            //TODO:            log.withoutCustomerData().error("Failed to label a connection", e);
-
+            connection.getClientInfo().setProperty(AURORA_REPLICA_ID, replicaId);
+        } catch (SQLException exception) {
+            throw new RuntimeException("Failed to label a replica connection with replica id: " + replicaId, exception);
```
wyrzyk, line 14: It will be easier to monitor it on prod if we use our own Exceptions.

src/test/java/com/atlassian/db/replica/it/example/aurora/AuroraClusterTest.java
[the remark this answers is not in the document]
wyrzyk: > It's not a target solution

Any not target solutions should live outside of the API.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -0,0 +1,55 @@
...
+            this.replicaConsistency = replicaConsistency;
+            return this;
+        }
+
+        public Builder auroraConnectionDetails(AuroraConnectionDetails auroraConnectionDetails) {
+            this.auroraConnectionDetails = auroraConnectionDetails;
+            return this;
+        }
+
+        public AuroraMultiReplicaConsistency build() {
```
wyrzyk, line 50: NIt: It can return the ReplicaConsistency interface. For example, it will allow us to switch implementations when rolling out new changes (with compatibility flag) instead of adding if's in one implementation.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -1,66 +1,45 @@
...
-                    return DriverManager.getConnection(url.toString(), props);
-                }
-            }).collect(toList());
-        } catch (SQLException throwables) {
-            throw new RuntimeException("TODO");
+            return discoverer.fetchReplicasUrls(connection).stream()
+                .map(auroraUrl -> new AuroraReplicaNode(auroraUrl, auroraConnectionDetails))
+                .collect(toList());
+        } catch (SQLException exception) {
+            throw new ReadReplicaDiscoveryOperationException(exception);
```
wyrzyk, line 28: Nice!

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
```diff
@@ -1,7 +1,6 @@
 package com.atlassian.db.replica.api;
 
 import com.atlassian.db.replica.internal.aurora.AuroraClusterDiscovery;
```
wyrzyk, line 3: Is the internal import still needed?

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -0,0 +1,44 @@
+package com.atlassian.db.replica.api;
+
+public final class AuroraConnectionDetails {
```
wyrzyk, line 3: Using an interface can make it more testable 🤔.

wyrzyk (approved):

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
tciesla-atlassian: There is no logic in this class (basic data transfer object) so how interface will make it more testable?

wyrzyk: Yes. It will not make it more testable now. It may change if the API evolve in the future. It can stay as-is.

wyrzyk (approved):

PR #115, opened 2021-09-01
Fixed connection leak in the AuroraClusterDiscovery

tciesla-atlassian:


src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -26,7 +26,7 @@ public void write(Connection main) {
     @Override
     public boolean isConsistent(Supplier<Connection> replicaSupplier) {
         return cluster.getReplicas(replicaSupplier).stream()
-            .allMatch(replica -> replicaConsistency.isConsistent(() -> replica.getConnectionSupplier().get()));
+            .allMatch(replica -> replicaConsistency.isConsistent(replica.getConnectionSupplier()));
```
dwatl, line 29: This place is still leaking the connection right? Closing connection by ReplicaConsistency is not part of the contract

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -18,8 +18,7 @@ public AuroraClusterDiscovery(AuroraConnectionDetails auroraConnectionDetails) {
     }
 
     public Collection<Database> getReplicas(Supplier<Connection> connectionSupplier) {
-        try {
-            final Connection connection = connectionSupplier.get();
+        try (final Connection connection = connectionSupplier.get()) {
```
dwatl, line 21: Perhaps it's worth adding unit-test as this is an ugly bug, WDYT?

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
tciesla-atlassian: IMO if someone takes connection from supplier then it's in their responsibility to correctly call close operation.

dwatl: Maybe, maybe not. The point is it's still leaking the connections.
We should either close it here or make it explicit in Java docs that not closing the connections from the supplier will lead to connection leaks.

I'm fine with moving it to separate PR to iterate faster. :)

dwatl (approved):

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
afaruga-atlassian: > in Java docs that not closing the connections from the supplier will lead to connection leaks.

like it

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
afaruga-atlassian: Ok, ` try-with-resources `
Is there option that connection will be a null ?

afaruga-atlassian: I'm not sure whether connection should be closed here, especially that you just state that it's ConnectionSupplier responsibility, how about doing a fix in product?

What if the connection will try to be reused? Than it wont work with product.

dwatl: Probably whoever is using supplier should close the connection, however, the supplier can return the same connection multiple times etc. We should have more explicit contracts with highlighted ownership to solve the problem :)

afaruga-atlassian: > We should have more explicit contracts with highlighted ownership to solve the problem :) 

yes, we should 🐛 
How we can measure the problem?

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
```diff
@@ -74,6 +74,19 @@ void shouldDiscoverReplicaConnectionsWhenScaledDown() {
         assertThat(replicas).hasSize(1);
     }
 
+    @Test
+    void shouldCloseDatabaseConnectionAfterDiscoveryReplicaConnections() throws SQLException {
```
wyrzyk, line 78: Interesting. I think it shouldn't. DualConnection closes the connection if the replica is not consistent, and keeps it if consistent. I think closing the connection breaks a hidden contract. It also shouldn't be a problem for a Waiter as it also closes the connection.

wyrzyk: Oh. I see ReplicaConsistency doesn't close the connection. AuroraClusterDiscovery does it. Seems to be 🆗.

afaruga-atlassian: Would be great to verify all of those in tests.

PR #118, opened 2021-09-02
Prepare for external connection-per-replica provider

pjaronski-atlassian:


src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -12,10 +15,11 @@
 
     private AuroraMultiReplicaConsistency(
         ReplicaConsistency replicaConsistency,
-        AuroraConnectionDetails auroraConnectionDetails
+        AuroraConnectionDetails auroraConnectionDetails,
+        BiFunction<String, Properties, AuroraConnectionSupplier> connectionPerReplicaProvider
```
wyrzyk, line 19: Can we avoid BiFunction in the API? It's hard to use, and it blocks future API evolution.

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
```diff
@@ -6,21 +6,7 @@
...
-import java.sql.NClob;
-import java.sql.PreparedStatement;
-import java.sql.SQLClientInfoException;
-import java.sql.SQLException;
-import java.sql.SQLWarning;
-import java.sql.SQLXML;
-import java.sql.Savepoint;
-import java.sql.Statement;
-import java.sql.Struct;
+import java.sql.*;
```
pjaronski-atlassian, line 9: that was not intended, I will revert this and make sure refactoring in my ide does not do this in the future

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
```diff
@@ -47,8 +33,7 @@ public void before() throws SQLException {
             "jdbc:postgresql://" + readerEndpoint + "/" + databaseName
         );
         auroraClusterDiscovery = new AuroraClusterDiscovery(
-            anAuroraConnectionDetailsBuilder().username(jdbcUsername).password(jdbcPassword).build()
-        );
+            anAuroraConnectionDetailsBuilder().username(jdbcUsername).password(jdbcPassword).build(), null);
```
wyrzyk, line 36: Why null? Not the test will not test/show new API usage.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -6,22 +6,34 @@
...
 
 import static java.util.stream.Collectors.toList;
 
 public final class AuroraClusterDiscovery {
     private final AuroraConnectionDetails auroraConnectionDetails;
+    private final BiFunction<String, Properties, AuroraConnectionSupplier> connectionPerReplicaProvider;
 
-    public AuroraClusterDiscovery(AuroraConnectionDetails auroraConnectionDetails) {
+    public AuroraClusterDiscovery(
+        AuroraConnectionDetails auroraConnectionDetails,
```
wyrzyk, line 21: We should not need AuroraConnectionDetails anymore.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -47,8 +52,17 @@ public Builder auroraConnectionDetails(AuroraConnectionDetails auroraConnectionD
             return this;
         }
 
+        public Builder connectionPerReplicaProvider(BiFunction<String, Properties, AuroraConnectionSupplier> connectionPerReplicaProvider) {
```
wyrzyk, line 55: You can mark `public Builder auroraConnectionDetails(AuroraConnectionDetails auroraConnectionDetails) {` as deprecated and transform AuroraConnectionDetails into connectionPerReplicaProvider here.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraConnectionSupplier.java
```diff
@@ -0,0 +1,9 @@
+package com.atlassian.db.replica.internal.aurora;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+@FunctionalInterface
+public interface AuroraConnectionSupplier {
+    Connection get() throws SQLException;
```
dwatl, line 8: We should specify replica here (it's important to connect to specific replica)

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraConnectionSupplier.java
```diff
@@ -0,0 +1,9 @@
+package com.atlassian.db.replica.internal.aurora;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+@FunctionalInterface
+public interface AuroraConnectionSupplier {
```
dwatl, line 7: Shouldn't this be part of SPI?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
```diff
@@ -32,20 +36,24 @@ public String getId() {
     public Supplier<Connection> getConnectionSupplier() {
         return () -> {
             try {
+                Connection connection = getAuroraConnectionSupplier().get();
                 return replicaNode.mark(
```
dwatl, line 40: I wonder if we can stop marking the connection if we'll have API to acquire connection to specific replica

dwatl (changes requested):

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: extracted the function into a separate interface

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
pjaronski-atlassian: provided an example usage of the new spi

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
pjaronski-atlassian: removed

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: done

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
```diff
@@ -32,20 +36,24 @@ public String getId() {
...
-        Properties properties = new Properties();
-        properties.setProperty("user", auroraConnectionDetails.getUsername());
-        properties.setProperty("password", auroraConnectionDetails.getPassword());
-        return DriverManager.getConnection(url.toString(), properties);
+    private AuroraConnectionSupplier getAuroraConnectionSupplier() {
+        if (connectionSupplier == null) {
+            Properties properties = new Properties();
+            properties.setProperty("user", auroraConnectionDetails.getUsername());
+            properties.setProperty("password", auroraConnectionDetails.getPassword());
+            connectionSupplier = connectionPerReplicaProvider.apply(auroraUrl.toString(), properties);
```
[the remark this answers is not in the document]
pjaronski-atlassian: done

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraConnectionSupplier.java
pjaronski-atlassian: renamed the interface to fit better and moved it to the spi package

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraConnectionSupplier.java
pjaronski-atlassian: we get connection per replica on a higher level, this is a secondary interface (outdated anyway)

src/main/java/com/atlassian/db/replica/api/ThrottledCache.java
```diff
@@ -62,7 +62,7 @@ private ThrottledCacheWithTimeout(Clock clock, Duration timeout) {
         @Override
         public Optional<T> get(Supplier<T> supplier) {
             maybeRefresh(supplier);
-            return Optional.ofNullable(lockValuePair.get().getValue());
+            return get();
```
pjaronski-atlassian, line 65: get() returns exactly Optional.ofNullable(lockValuePair.get().getValue())
DRY?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraConnectionSupplier.java
pjaronski-atlassian, line 7: outdated; will remove in the next commit

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
pjaronski-atlassian: replicaNode is not being used anywhere else in this class so marking the node does virtually nothing. it could be simply removed from the code with no effect on this class behaviour

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
```diff
@@ -46,9 +46,12 @@ public void before() throws SQLException {
...
         );
-        auroraClusterDiscovery = new AuroraClusterDiscovery(
-            anAuroraConnectionDetailsBuilder().username(jdbcUsername).password(jdbcPassword).build()
-        );
+        auroraClusterDiscovery = new AuroraClusterDiscovery((url) -> () -> {
+            Properties properties = new Properties();
+            properties.setProperty("user", jdbcUsername);
+            properties.setProperty("password", jdbcPassword);
+            return DriverManager.getConnection(url, properties);
+        });
```
pjaronski-atlassian, line 54: simple usage of new spi interfaces. creates and returns a new connection every time it is requested

src/main/java/com/atlassian/db/replica/spi/ConnectionPoolPerUrlProvider.java
```diff
@@ -0,0 +1,6 @@
+package com.atlassian.db.replica.spi;
+
+@FunctionalInterface
+public interface ConnectionPoolPerUrlProvider {
+    ConnectionPool getConnectionPool(String url);
```
wyrzyk, line 5: Suppose it's per URL then java.net.URL instead of String. Or maybe it's not (doesn't have to be) URL, but id. String id would be more generic.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -10,12 +10,13 @@
     private final ReplicaConsistency replicaConsistency;
     private final AuroraClusterDiscovery cluster;
 
-    private AuroraMultiReplicaConsistency(
-        ReplicaConsistency replicaConsistency,
-        AuroraConnectionDetails auroraConnectionDetails
-    ) {
+    public AuroraMultiReplicaConsistency(ReplicaConsistency replicaConsistency, AuroraClusterDiscovery cluster) {
```
wyrzyk, line 13: `AuroraClusterDiscovery` is internal and leaks to the API (you need an instance of AuroraClusterDiscovery to use AuroraMultiReplicaConsistency).

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -1,44 +0,0 @@
-package com.atlassian.db.replica.api;
-
-public final class AuroraConnectionDetails {
```
wyrzyk, line 3: Removing API class breaks compatibility.

wyrzyk (changes requested):
Please fix the API issues

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
wyrzyk: https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_Classes

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: removed the leak

src/main/java/com/atlassian/db/replica/spi/ConnectionPoolPerUrlProvider.java
pjaronski-atlassian: - IMO introducing java.net.URL is unnecessary complexity
- url clearly indicates what has to be passed into the method; id might be confusing since node id is not necessarily equal to node url

I could rename to `-PerReplica-` instead of `-PerUrl-` if that would increase the readability?

wyrzyk: > IMO introducing java.net.URL is unnecessary complexity

We have a type to represent URI build in the language. Why increasing type safety is more complex? Using Objects instead of string would be even less complex?

It's a string, so any string can be passed. 

> id might be confusing since node id is not necessarily equal to node URL

It doesn't have to be node id.

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
pjaronski-atlassian: restored the class to preserve compatibility

src/main/java/com/atlassian/db/replica/spi/ConnectionPoolPerUrlProvider.java
pjaronski-atlassian: - introducing java.net.URL will require us to handle all the exceptions (simplest case scenario we just rethrow them or do not catch them at all).
- jdbc drivers already handle the url string when creating database connections. handling the url on our side would be just duplicate work
- jdbc drivers use different protocols than java.net.URL expects. e.g. java.net.URL does not recognise the "jdbc:postgresql" protocol and raises a `java.net.MalformedURLException` while trying to instantiate the class. Can you provide a way to transform a database String url into a java.net.URL?

wyrzyk: > introducing java.net.URL will require us to handle all the exceptions (simplest case scenario we just rethrow them or do not catch them at all).

URI/URL have a specified syntax, and the type ensures it's what it should be. It gives us safety for free (fail fast). Other options are to validate the syntax on our own on pass it on. It may be harder to debug and understand why something is failing.

> jdbc drivers already handle the url string when creating database connections.

JDBC is a very stable API. It will contain lots of legacy stuff. But we don't need to inherit it. We can be more type-safe than JDBC. 

> jdbc drivers use different protocols than java.net.URL

Use URI.

pjaronski-atlassian: > URI/URL have a specified syntax, and the type ensures it's what it should be

But the valid syntax for a database url is only a subset of what is permissible by the URI class. e.g: `http://localhost:1234/index.html` is a valid URI syntax, but it is not a valid database url syntax. 

> It gives us safety for free (fail fast)

URI does not guarantee that the underneath string is a valid database url, so this is not fail-fast. 
Neither is it free because we would be paying the overhead of parsing a string into a URI.

If we want to pass a database url in a type-safe manner, then we should write our own class that checks for the validity of a database url.

wyrzyk: Yes. URI is more type-safe than String but could be more strict if we introduce our own type.

pjaronski-atlassian: would `AuroraJdbcUrl` be a good candidate for this?

pjaronski-atlassian: obviously, it would have to be in the api package and be aurora-agnostic but it's just a suggestion

wyrzyk: > JdbcUrl) and Aurora-specific implementation (URI is also more generic).

We can also use URI with the partial type safety and let the implementation assert on DB specific aspects.

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcProtocol.java
```diff
@@ -0,0 +1,17 @@
+package com.atlassian.db.replica.spi.jdbc;
```
wyrzyk, line 1: Is it SPI or API?

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -36,9 +56,17 @@ public Builder password(String password) {
             return this;
         }
 
+        /**
+         * @return new instance of this class
```
tciesla-atlassian, line 60: comment "@return new instance of this class" doesn't add any value, IMO it make code more bloated

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -31,25 +36,39 @@ public boolean isConsistent(Supplier<Connection> replicaSupplier) {
...
-        private AuroraConnectionDetails auroraConnectionDetails;
+        private ConnectionPoolPerUrlProvider connectionPoolPerUrlProvider;
 
+        public Builder replicaConsistency(ReplicaConsistency replicaConsistency) {
+            this.replicaConsistency = replicaConsistency;
+            return this;
+        }
+
+        /**
+         * @return new instance of this class
```
tciesla-atlassian, line 47: similar as above, https://github.com/atlassian-labs/db-replica/pull/118/files#r702889777

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcProtocol.java
```diff
@@ -0,0 +1,17 @@
...
+
+public enum JdbcProtocol {
```
wyrzyk, line 3: Enums are hard to evolve (https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#255-do-not-add-new-values-to-a-released-enumeration). 

It should not be an issue, as enums should not evolve (https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html).

> An enum type is a special data type that enables for a variable to be a set of predefined constants. 

But it seems we don't have a full list of current and the future JdbcProtocols here 😉.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -11,17 +12,21 @@
...
     }
 
     public Collection<Database> getReplicas(Supplier<Connection> connectionSupplier) {
         try (final Connection connection = connectionSupplier.get()) {
             final AuroraReplicasDiscoverer discoverer = createDiscoverer(connection);
             return discoverer.fetchReplicasUrls(connection).stream()
-                .map(auroraUrl -> new AuroraReplicaNode(auroraUrl, auroraConnectionDetails))
+                .map(auroraUrl -> {
+                    String id = auroraUrl.getEndpoint().getServerId();
+                    ConnectionPool connectionPool = connectionPoolPerUrlProvider.getConnectionPool(auroraUrl);
```
tciesla-atlassian, line 27: Here new connection pool is created every time we want to fetch lsn, connection pool caching will be implemented in separate PR?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
```diff
@@ -1,51 +1,37 @@
...
 
     @Override
     public Supplier<Connection> getConnectionSupplier() {
         return () -> {
             try {
-                return replicaNode.mark(
-                    getConnection(auroraUrl),
-                    auroraUrl.getEndpoint().getServerId()
-                );
+                return connectionPool.getConnection();
```
tciesla-atlassian, line 31: What about marking connection? It will be done on Jira side? (`replicaNode.mark`)

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcUrl.java
```diff
@@ -0,0 +1,56 @@
+package com.atlassian.db.replica.spi.jdbc;
+
+public class JdbcUrl {
```
wyrzyk, line 3: Did you think about the changes with Occam's razor in mind? I mean going from String to URI doesn't increase the number of entities, but going further and introducing three entitles just to pass the URI seems to be overkill (especially that it could be handled without the new API/SPI. ConnectionPoolPerReplicaProvider implementation can provide validation specific to the implemented database).

src/main/java/com/atlassian/db/replica/spi/SuppliedCache.java
```diff
@@ -14,9 +14,7 @@
     Optional<T> get(Supplier<T> supplier);
 
     /**
-     * Returns the cached value
-     *
-     * @return
+     * @return the cached value
```
tciesla-atlassian, line 17: isn't it obvious? :)

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcUrl.java
```diff
@@ -0,0 +1,56 @@
+package com.atlassian.db.replica.spi.jdbc;
```
wyrzyk, line 1: Is it SPI or API?

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcUrl.java
```diff
@@ -0,0 +1,56 @@
...
+        public Builder database(String database) {
+            this.database = database;
+            return this;
+        }
+
+        public JdbcUrl build() {
+            return new JdbcUrl(protocol, endpoint, database);
+        }
+
+        public String toString() {
```
tciesla-atlassian, line 52: do we need toString in the builder?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraJdbcUrl.java
```diff
@@ -1,12 +1,18 @@
 package com.atlassian.db.replica.internal.aurora;
 
-public class AuroraJdbcUrl {
-    private static final String PREFIX = "jdbc:postgresql://";
+import com.atlassian.db.replica.spi.jdbc.JdbcProtocol;
+import com.atlassian.db.replica.spi.jdbc.JdbcUrl;
 
+public class AuroraJdbcUrl extends JdbcUrl {
```
wyrzyk, line 6: It's internal. I guess we don't want to call it from the product codebase.

wyrzyk: URI gives us stricter validation of the input than the current implementation. Is it intentional?

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -31,25 +36,39 @@ public boolean isConsistent(Supplier<Connection> replicaSupplier) {
...
+         * @deprecated use {@link AuroraMultiReplicaConsistency#builder()} instead.
+         */
+        @Deprecated
         public static Builder anAuroraMultiReplicaConsistencyBuilder() {
             return new Builder();
         }
 
-        public Builder replicaConsistency(ReplicaConsistency replicaConsistency) {
-            this.replicaConsistency = replicaConsistency;
+        public Builder connectionPoolPerUrlProvider(ConnectionPoolPerUrlProvider connectionPoolPerUrlProvider) {
```
tciesla-atlassian, line 55: IMO we should rename type or at least parameter. Shouldn't client know that this have to be connection pool provider for read replicas? Configuration between main / replica connection pool may differ, so we have to suggest that we will be creating connections to replicas.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -31,25 +36,39 @@ public boolean isConsistent(Supplier<Connection> replicaSupplier) {
...
+            this.connectionPoolPerUrlProvider = connectionPoolPerUrlProvider;
             return this;
         }
 
+        /**
+         * @param auroraConnectionDetails details
+         * @return this builder
```
tciesla-atlassian, line 62: unnecessary comments:
```
* @param auroraConnectionDetails details
* @return this builder
```

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
pjaronski-atlassian: will remove the unnecessary doc

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: will remove as well

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: will remove

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: will rename to -PerReplica-

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
pjaronski-atlassian: it's `getConnectionPool` not `createConnectionPool`, it is client's responsibility to reuse existing connection pool

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
pjaronski-atlassian: it was removed per @dwatl's suggestion

src/main/java/com/atlassian/db/replica/spi/SuppliedCache.java
pjaronski-atlassian: yes, I just moved the obvious comment to fit into the javadoc format

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcUrl.java
wyrzyk: Does it matter? 😉

pjaronski-atlassian: other builders in the repository do not override the `toString` method so I guess I could remove it for consistency's sake.
otoh I think it is a good practice to always provide the `toString` implementation as the default one is very poor

wyrzyk: Yes. The default implementation is useless. I think @tciesla-atlassian meant there's a high chance "You aren't gonna need it".

pjaronski-atlassian: true, will remove it then

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
```diff
@@ -10,22 +18,34 @@ private AuroraConnectionDetails(String username, String password) {
...
+        return this.username;
     }
 
     public String getPassword() {
-        return password;
+        return this.password;
+    }
+
+    public ConnectionPoolPerUrlProvider convert() {
+        return url -> () -> DriverManager.getConnection(
```
afaruga-atlassian, line 34:
```
DriverManager.getConnection(
            url.toString(),
            username,
            password
```
This could be a default implementation of `ConnectionPoolPerUrlProvider`

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcUrl.java
pjaronski-atlassian: It should be `api` because we do not require the client to implement/extend this.
Will move it to the appropriate package.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -11,17 +12,21 @@
 import static java.util.stream.Collectors.toList;
 
 public final class AuroraClusterDiscovery {
-    private final AuroraConnectionDetails auroraConnectionDetails;
+    private final ConnectionPoolPerUrlProvider connectionPoolPerUrlProvider;
 
-    public AuroraClusterDiscovery(AuroraConnectionDetails auroraConnectionDetails) {
-        this.auroraConnectionDetails = auroraConnectionDetails;
+    public AuroraClusterDiscovery(ConnectionPoolPerUrlProvider connectionPoolPerUrlProvider) {
```
afaruga-atlassian, line 17: I think its better to extract `Supplier<Connection> getConnectionSupplier` instead of 
`ConnectionPoolPerUrlProvider`
Why ? We shouldn't play with `connectionPool` in `AuroraClusterDiscovery`  this is to greedy IMHO. 
AuroraClusterDiscovery just want to get connection for URL
`ConnectionUrlProvider`

afaruga-atlassian: It will give more control on product side around connectionPool and connection, and I think this is what we want to achieve yeah ?

afaruga-atlassian: another advantage is 1 less SPI to maintain `com.atlassian.db.replica.spi.ConnectionPool`

src/test/java/com/atlassian/db/replica/api/AuroraClusterDiscoveryTest.java
```diff
@@ -46,9 +31,11 @@ public void before() throws SQLException {
             auroraClusterMock.getMainConnection(),
             "jdbc:postgresql://" + readerEndpoint + "/" + databaseName
         );
-        auroraClusterDiscovery = new AuroraClusterDiscovery(
-            anAuroraConnectionDetailsBuilder().username(jdbcUsername).password(jdbcPassword).build()
-        );
+        auroraClusterDiscovery = new AuroraClusterDiscovery(url -> () -> DriverManager.getConnection(
```
afaruga-atlassian, line 34: Even in test `ConnectionPool` isn't tested, only `getConnection`

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraJdbcUrl.java
pjaronski-atlassian: will use URI's validation when constructing JdbcUrl

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/AuroraConnectionProvider.java
```diff
@@ -32,9 +31,10 @@ public Connection getReplicaConnection() throws SQLException {
...
 
     private Connection getConnection(String url) throws SQLException {
-        final Properties props = new Properties();
-        props.setProperty("user", "postgres");
-        props.setProperty("password", System.getenv("password"));
-        return DriverManager.getConnection(url, props);
+        return DriverManager.getConnection(
+            url,
+            "postgres",
+            System.getenv("password")
```
afaruga-atlassian, line 37: this is strange. I think product should determine the way pass password
It also complicate testing

afaruga-atlassian: Ah sorry 🤦 this is test implemention. All good

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/api/SequenceReplicaConsistency.java
```diff
@@ -26,21 +25,16 @@
     private final ConcurrentHashMap<String, SuppliedCache<Long>> multiReplicaLsnCache;
     private final ReplicaNode replicaNode;
 
-    public static final class Builder {
```
afaruga-atlassian, line 29: Maintaining this test, while it can be used because of none Aurora infra, it's an additional cost, without benefits hopefully soon we will have it.

src/main/java/com/atlassian/db/replica/spi/ConnectionPoolPerUrlProvider.java
pjaronski-atlassian: Introduced a JdbcUrl class that's a wrapper for a database url string.
Added a Builder class for manually constructing a valid database url.
Re-used existing validity checks of the `java.net.URI` class

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcProtocol.java
pjaronski-atlassian: It's API since we do not require the client to implement/extend this.
Will move it to the `api` package along with the `JdbcUrl` class

src/main/java/com/atlassian/db/replica/spi/jdbc/JdbcProtocol.java
pjaronski-atlassian: replaced enum with final class with a private constructor and static instances

src/main/java/com/atlassian/db/replica/api/AuroraConnectionDetails.java
pjaronski-atlassian: having this as a default implementation would be misleading since we would not be actually managing any connection pools but creating a new connection every time

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/api/SequenceReplicaConsistency.java
dwatl: I think it's work in progress

src/main/java/com/atlassian/db/replica/spi/ConnectionPoolPerReplicaProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.db.replica.spi;
+
+import com.atlassian.db.replica.api.jdbc.JdbcUrl;
+
+@FunctionalInterface
+public interface ConnectionPoolPerReplicaProvider {
```
dwatl, line 6: Do we really need to know about the connection pool? We can simplify all of that by:
ReplicaUrl -> Connection

Pools, caches are implementation detail IMO

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -11,17 +12,21 @@
...
 
     public Collection<Database> getReplicas(Supplier<Connection> connectionSupplier) {
         try (final Connection connection = connectionSupplier.get()) {
             final AuroraReplicasDiscoverer discoverer = createDiscoverer(connection);
             return discoverer.fetchReplicasUrls(connection).stream()
-                .map(auroraUrl -> new AuroraReplicaNode(auroraUrl, auroraConnectionDetails))
+                .map(auroraUrl -> {
+                    String id = auroraUrl.getEndpoint().getServerId();
+                    ConnectionPool connectionPool = connectionPoolPerReplicaProvider.getConnectionPool(auroraUrl);
+                    return new AuroraReplicaNode(id, connectionPool);
```
dwatl, line 28: What is the responsibility of AuroraReplicaNode now? How is it different than ConnectionPool

src/test/java/com/atlassian/db/replica/it/example/aurora/replica/api/SequenceReplicaConsistency.java
afaruga-atlassian: Yeah, 🤞 🤞 🤞 🤞

afaruga-atlassian: I'm totally for it 🚀

src/main/java/com/atlassian/db/replica/api/jdbc/JdbcUrl.java
```diff
@@ -0,0 +1,51 @@
+package com.atlassian.db.replica.api.jdbc;
+
+import java.net.URI;
+
+public class JdbcUrl {
```
wyrzyk, line 5: Please make it final.

src/main/java/com/atlassian/db/replica/api/jdbc/JdbcUrl.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final String PREFIX = "jdbc";
+
+    private final String internalUrl;
+
+    public JdbcUrl(JdbcProtocol protocol, String endpoint, String database) {
```
wyrzyk, line 10: We have a builder. The constructor doesn't have to be public.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraJdbcUrl.java
```diff
@@ -1,12 +1,18 @@
 package com.atlassian.db.replica.internal.aurora;
 
-public class AuroraJdbcUrl {
-    private static final String PREFIX = "jdbc:postgresql://";
+import com.atlassian.db.replica.api.jdbc.JdbcProtocol;
+import com.atlassian.db.replica.api.jdbc.JdbcUrl;
 
+public class AuroraJdbcUrl extends JdbcUrl {
```
wyrzyk, line 6: Why not composition? Making JdbcUrl not final will make it hard to evolve.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
```diff
@@ -1,51 +1,37 @@
...
-                    auroraUrl.getEndpoint().getServerId()
-                );
+                return replicaConnectionProvider.getReplicaConnection();
             } catch (SQLException exception) {
                 throw new ReadReplicaConnectionCreationException(exception);
             }
         };
     }
-
-    private Connection getConnection(AuroraJdbcUrl url) throws SQLException {
```
wyrzyk, line 45: 👍

src/main/java/com/atlassian/db/replica/spi/ConnectionProvider.java
```diff
@@ -3,17 +3,12 @@
...
     /**
-     * @return a connection to the main databsase
+     * @return a connection to the main database
      */
     Connection getMainConnection() throws SQLException;
-
-    /**
-     * @return a connection to a replica database
-     */
-    Connection getReplicaConnection() throws SQLException;
```
wyrzyk, line 18: The change is not compatible.

wyrzyk: I see ReplicaConnectionProvider takes it.

wyrzyk: What's the reason to change it now?

src/test/java/com/atlassian/db/replica/spi/DefaultReplicaConnectionPerUrlProvider.java
```diff
@@ -0,0 +1,20 @@
+package com.atlassian.db.replica.spi;
```
wyrzyk, line 1: Is it SPI or API?

src/test/java/com/atlassian/db/replica/spi/DefaultReplicaConnectionPerUrlProvider.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.atlassian.db.replica.api.jdbc.JdbcUrl;
+
+import java.sql.DriverManager;
+
+public class DefaultReplicaConnectionPerUrlProvider implements ReplicaConnectionPerUrlProvider {
```
wyrzyk, line 7: Do we need this non-working in prod default?

wyrzyk: I see it's not API it's for the tests. 👌

src/main/java/com/atlassian/db/replica/api/jdbc/JdbcUrl.java
pjaronski-atlassian: ok

src/main/java/com/atlassian/db/replica/api/jdbc/JdbcUrl.java
pjaronski-atlassian: ok

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraJdbcUrl.java
pjaronski-atlassian: made JdbcUrl final and introduced a `toJdbcUrl` method to create a JdbcUrl using an AuroraJdbcUrl

src/main/java/com/atlassian/db/replica/spi/ConnectionProvider.java
pjaronski-atlassian: I extracted this method into `ReplicaConnectionProvider` interface to reuse it in `ReplicaConnectionPerUrlProvider` because as per @dwatl's suggestion we shouldn't make the lib connection-pool-aware

src/test/java/com/atlassian/db/replica/spi/DefaultReplicaConnectionPerUrlProvider.java
pjaronski-atlassian: It's a naive implementation used only in tests so I guess it's neither `spi` nor `api` (maybe `internal` then?).
I could make it a default implementation and move it to the `src/main/.../api` package
@wyrzyk WDYT?

wyrzyk: Make it internal please.

pjaronski-atlassian: > I could make it a default implementation

Never mind, I just saw the next comment (about this implementation not working in prod). 
I will move it to `src/test/.../internal` then

wyrzyk (dismissed):

src/main/java/com/atlassian/db/replica/spi/ReplicaConnectionProvider.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+@FunctionalInterface
+public interface ReplicaConnectionProvider {
+    /**
+     * @return a connection to a replica database
+     */
+    Connection getReplicaConnection() throws SQLException;
```
dwatl, line 11: Why this is separated from ReplicaConnectionPerUrlProvider?

wyrzyk: I think it's a bad idea to merge it. Most of the use cases (all until now) didn't require providing a way to connect with a replica via URL.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicaNode.java
```diff
@@ -1,51 +1,37 @@
...
 import com.atlassian.db.replica.internal.Database;
+import com.atlassian.db.replica.spi.ReplicaConnectionProvider;
 
 import java.sql.Connection;
-import java.sql.DriverManager;
 import java.sql.SQLException;
-import java.util.Properties;
 import java.util.function.Supplier;
 
 public class AuroraReplicaNode implements Database {
```
dwatl, line 10: I feel like this one could be removed

dwatl (approved):

src/main/java/com/atlassian/db/replica/spi/ReplicaConnectionProvider.java
wyrzyk: We also require ConnectionProvider to be implemented to use DualConnection, so the change would:
- increase the entry threshold
- break the API (I guess we are not able to provide a default implementation that would work for all the implementations in a compatible way).

wyrzyk (approved):

afaruga-atlassian (approved):
Decent work 🚀

PR #120, opened 2021-09-07
Added cache to AuroraClusterDiscovery

tciesla-atlassian:


src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -47,8 +55,17 @@ public Builder auroraConnectionDetails(AuroraConnectionDetails auroraConnectionD
             return this;
         }
 
+        public Builder discoveredReplicasCache(SuppliedCache<Collection<Database>> discoveredReplicasCache) {
```
wyrzyk, line 58: It seems to conflict with (https://github.com/atlassian-labs/db-replica/pull/118). CC: @pjaronski-atlassian

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
```diff
@@ -12,12 +14,22 @@
...
+    private AuroraClusterDiscovery(
+        AuroraConnectionDetails auroraConnectionDetails,
+        SuppliedCache<Collection<Database>> discoveredReplicasCache
+    ) {
         this.auroraConnectionDetails = auroraConnectionDetails;
+        this.discoveredReplicasCache = discoveredReplicasCache;
     }
 
     public Collection<Database> getReplicas(Supplier<Connection> connectionSupplier) {
+        return discoveredReplicasCache.get(() -> fetchReplicas(connectionSupplier))
```
wyrzyk, line 28: fetchReplicas if it's null then fetch it again?

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -32,6 +39,7 @@ public boolean isConsistent(Supplier<Connection> replicaSupplier) {
     public static final class Builder {
         private ReplicaConsistency replicaConsistency;
         private AuroraConnectionDetails auroraConnectionDetails;
+        private SuppliedCache<Collection<Database>> discoveredReplicasCache = new NoCacheSuppliedCache<>();
```
wyrzyk, line 42: What's the cache for?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
tciesla-atlassian: yeah, when there will be no value in the cache'a then we need to fetch them. We cannot return empty collection here because it will be used in consistency checking -> empty replicas collection means that we are consistent with no replicas... which is incorrect.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
tciesla-atlassian: We don't need to fetch existing replicas every millisecond and in parallel, but with certain throttling, like every 100ms or maybe 1s. Replicas are scaled up and down dynamically and we have to balance between how fast we discover the change vs what overhead we are adding.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
wyrzyk: Yes, but why do we call fetchReplicas(connectionSupplier) twice and expect different results?

```discoveredReplicasCache.get(() -> fetchReplicas(connection supplier))``` loads the cache from the supplier if the cache is empty. If the cache load fails, then we call the supplier again and return the results without caching it.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
wyrzyk: Can you add a JavaDoc?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
tciesla-atlassian: If cache loads data from supplier when is empty, then why `.get(supplier)` returns Optional? I guess it may only happen in case of some error. WDYT should i create custom exception here?

wyrzyk: I don't know why it uses optional.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
tciesla-atlassian: It's not in conflict with @pjaronski-atlassian changes. Piotr replaces AuroraConnectionDetails with connection pool provider. I'm adding additional cache for discoverer.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
tciesla-atlassian: Ok, I see good solution. I will return empty collection and delegate to the client (consistency checking) decision. I guess for empty collection `isConsistent` should return false (we are not consistent with replicas where there are no replicas).

pjaronski-atlassian (approved):

wyrzyk (approved):

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
dwatl: If we have 0 replicas then we are always consistent, no?

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: `Database` is internal. Isn't this an implementation detail leak?

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraClusterDiscovery.java
pjaronski-atlassian: I also think that an empty collection of replicas should indicate that we are consistent - not the other way around

tciesla-atlassian: IMO we shouldn't call `isConsistent` when there are no replicas. When we check replicas consistency and there are no replicas for me it sounds like error. At the end it doesn't matter because we forward move to main when replicas are unavailable/inconsistent.

wyrzyk: > When we check replicas consistency and there are no replicas for me it sounds like error.
Why error? If all replicas fail then there will be some time without any replica (especially for the single replica setups).

empty collection = no replicas
exception = can't check consistency (by default DualConnection switches to the main database)

PR #122, opened 2021-09-08
mark release 2.2.6

tciesla-atlassian:


CHANGELOG.md
```diff
@@ -22,7 +22,13 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-2.2.4...master
+[Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-2.2.6...master
+
+## [2.2.6] - 2021-09-08
+[2.2.6]: https://github.com/atlassian-labs/db-replica/compare/release-2.2.4...release-2.2.6
+
+### Fixed
+- Additional caching in `AuroraClusterDiscovery`
```
wyrzyk, line 31: Shouldn't be AuroraMultiReplicaConsistency? `AuroraClusterDiscovery` is an internal.

CHANGELOG.md
```diff
@@ -22,7 +22,13 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-2.2.4...master
+[Unreleased]: https://github.com/atlassian-labs/db-replica/compare/release-2.2.6...master
+
+## [2.2.6] - 2021-09-08
+[2.2.6]: https://github.com/atlassian-labs/db-replica/compare/release-2.2.4...release-2.2.6
+
+### Fixed
+- Additional caching in `AuroraMultiReplicaConsistency`
```
wyrzyk, line 31: I'm not sure if it's already released, but we probably should not release it as @pjaronski-atlassian noticed it may force library consumers to use internals.

tciesla-atlassian: This class is already in the api and it's released. Without release it will block current rollout.

wyrzyk: > Without release it will block current rollout.

Blocking rollout seems to be less important than breaking API.

> This class is already in the api and it's released.

Let's fix it ASAP (move the class to API).

wyrzyk (approved):

CHANGELOG.md
dwatl: @wyrzyk WDYT about milestones/beta versions? We potentially could be releasing unstable versions hence have fewer API breaking changes.

wyrzyk: You can always release SNAPSHOT or anything.

PR #125, opened 2021-09-15
Autoclose fetched replica connections when checking consistency

pjaronski-atlassian:


build.gradle.kts
```diff
@@ -34,6 +34,7 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter")
     testImplementation("org.assertj:assertj-core:3.19.0")
     testImplementation("org.mockito:mockito-core:3.11.0")
+    testImplementation("org.mockito:mockito-junit-jupiter:3.11.0")
```
pjaronski-atlassian, line 37: allows using `@ExtendWith(MockitoExtension.class)`

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -39,12 +40,16 @@ public void write(Connection main) {
...
-            .allMatch(replica -> replicaConsistency.isConsistent(() -> replica.getConnectionSupplier().get()));
+            .map(replica -> {
+                try (Connection connection = replica.getConnectionSupplier().get()) {
+                    return replicaConsistency.isConsistent(() -> connection);
+                } catch (SQLException exception) {
+                    return false;
+                }
+            })
+            .reduce(Boolean::logicalAnd)
+            .orElse(false);
```
pjaronski-atlassian, line 52: if we want to autoclose obtained connections, then we cannot use `allMatch` as it is a short-circuit operator

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
```diff
@@ -39,12 +40,16 @@ public void write(Connection main) {
...
-            return false;
-        }
-
         return replicas.stream()
-            .allMatch(replica -> replicaConsistency.isConsistent(() -> replica.getConnectionSupplier().get()));
+            .map(replica -> {
+                try (Connection connection = replica.getConnectionSupplier().get()) {
+                    return replicaConsistency.isConsistent(() -> connection);
+                } catch (SQLException exception) {
+                    return false;
```
pjaronski-atlassian, line 48: I'm open to suggestions as to what should happen when this exception is caught

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
```diff
@@ -0,0 +1,103 @@
...
+            );
+            nodes.add(node);
+        }
+        when(discoveredReplicasCache.get(any())).thenReturn(Optional.of(nodes));
+
+        return replicas;
+    }
+
+    @Test
+    void shouldNotBeConsistentWhenNoReplicasAreAvailable() throws SQLException {
```
pjaronski-atlassian, line 60: `throws` is not needed here, will remove it with the next commit

wyrzyk, line 60: So when all replicas are down, and only the main database survived, I will never move forward.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
wyrzyk: We can not catch the exception, and allow the API caller to decide. 🤔

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
wyrzyk: Can you elaborate?
AllMatch will not obtain all connections, but should close all the obtained connections.

wyrzyk (approved):

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
dwatl: We should change it, no inconsistent replicas means we are consistent.

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
dwatl: IMO Re-throw as runtime exception

wyrzyk: Why RE instead of SQLE?

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: you're right, will refactor this

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
pjaronski-atlassian: will change it so that no replicas indicate that we are consistent

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
dwatl: Checked exceptions inside lambda are not allowed AFAIK

pjaronski-atlassian: we cannot rethrow as SQLE because be cannot change the signature of the `isConsistent` method

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Mock
+    private Connection supplierConnection;
+    @Mock
+    private SuppliedCache<Collection<Database>> discoveredReplicasCache;
+
+    private AuroraMultiReplicaConsistency sut;
+
+    private void mockReplicaConnections(int count) {
+        replicas = new LinkedList<>();
```
wyrzyk, line 33: Why reusing replicas between tests? The method could return a new copy the risk of flakiness will decrease.

wyrzyk: Does noone use it?

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
pjaronski-atlassian: unless we added it in the `ReplicaConsistency` interface as well

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
wyrzyk: I see it's shared between the test cases.

wyrzyk (approved):

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
pjaronski-atlassian: will make the `replicas` a local variable for each test

wyrzyk (dismissed):

src/main/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistency.java
dwatl: In the future we could return some useful object instead of a dummy boolean. It could contain all replicas with boolean/list of errors.

wyrzyk: @dwatl Remember `ReplicaConsistency` is SPI. DualConnection uses it to determine if the replica is consistent or not. Still, we'll need to decide what to do with the exception inside of the library. Currently, it's re-thrown.
```
try {
            isConsistent = consistency.isConsistent(() -> readConnection.get(decisionBuilder));
        } catch (Exception e) {
            throw e;
        }
```

> It could contain all replicas with boolean/list of errors.

Multi-replica lives only in one `ReplicaConsistency` implementation. Returning data specific only for a single implementation from API would make a poor abstraction.

src/main/java/com/atlassian/db/replica/api/exception/ConnectionCouldNotBeClosedException.java
```diff
@@ -0,0 +1,7 @@
+package com.atlassian.db.replica.api.exception;
+
+public class ConnectionCouldNotBeClosedException extends RuntimeException {
```
wyrzyk, line 3: Make it final.

wyrzyk (dismissed):

wyrzyk (approved):

wyrzyk:
falcon-go

dwatl (approved):
🤞

src/test/java/com/atlassian/db/replica/api/AuroraMultiReplicaConsistencyTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+class AuroraMultiReplicaConsistencyTest {
+    @Mock
+    private Connection supplierConnection;
+    @Mock
+    private SuppliedCache<Collection<Database>> discoveredReplicasCache;
+
+    private AuroraMultiReplicaConsistency sut;
```
afaruga-atlassian, line 28: sut ?
is is `system under test` ? 
I would prefer `underTest`  ;)

afaruga-atlassian (commented):
👍 🤞

PR #127, opened 2021-09-23
Run insert queries on the main database

wyrzyk:


dwatl (dismissed):

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -13,14 +13,18 @@ public SqlQuery(String sql) {
...
+        return sqlFunction.isFunctionCall(sql) || isUpdate() || isDelete() || isInsert();
     }
 
     public boolean isSelectForUpdate() {
         final String trimmedQuery = trimForSelectForUpdateCheck();
         return trimmedQuery.contains("for update") || trimmedQuery.contains("FOR UPDATE");
     }
 
+    public boolean isInsert() {
+        return sql.startsWith("insert") || sql.startsWith("INSERT");
```
afaruga-atlassian, line 25: `sql.contains("insert")`

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -54,6 +54,23 @@
...
 
+    @Test
+    public void shouldUseMainConnectionForInserts() throws SQLException {
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).build();
+
+        connection.prepareStatement("insert into \"public\".\"fvv\" (\"cd\", \"iid\", \"did\", \"s\")\n" +
```
afaruga-atlassian, line 65: Did that happens on production and results with failures from Replica ?
Asking cause I haven't see it yet.

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
wyrzyk: Most db-replicas decisions are heuristics focused on performance over accuracy. We update the rules when we see it doesn't work on prod. String#contains is O(nm) (for the Java implementations I've seen. It's possible to implement it better O(n), but it's still costly), and startsWith is O(1) and passes the case from prod.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: Yes. See the internal stand-up channel.

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
afaruga-atlassian: `Most db-replicas decisions are heuristics focused on performance over accuracy. `

👍  Ok, lets follow this

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: `"ERROR: cannot execute INSERT in a read-only transaction"`

I see

afaruga-atlassian (approved):

PR #137, opened 2021-12-20
Issue/pcs 66842 keep statement timeout while switching connections

wyrzyk:


src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1653,6 +1655,37 @@ public void shouldKeepUsingReplicaAfterSettingTimeout() throws SQLException {
...
+        final List<Statement> mainConnectionStatements = allStatements
+            .stream()
+            .filter(statement -> !replicaStatements.contains(statement))
+            .collect(Collectors.toList());
+        assertThat(connectionProvider.getProvidedConnectionTypes())
+            .containsExactly(REPLICA, MAIN);
+        assertThat(replicaStatements).hasSize(1);
+        verify(replicaStatements.get(0)).execute("set statement_timeout to 30000");
+        assertThat(mainConnectionStatements).hasSize(1);
+        verify(mainConnectionStatements.get(0)).execute("set statement_timeout to 30000");
```
afaruga-atlassian, line 1686: is there option to have different `set statement_timeout to 30000` only on MAIN ?

wyrzyk: `set` changes configuration for a current connection. From the perspective of DualConnection's user, there's no replica/main. There's only DualConnection. The user can't use different settings for main and replica as the internal connections are transparent and there's no dedicated API to set one or another.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1653,6 +1655,37 @@ public void shouldKeepUsingReplicaAfterSettingTimeout() throws SQLException {
             .containsOnly(REPLICA);
     }
 
+    @Test
```
afaruga-atlassian, line 1658: You were close in `shouldKeepUsingReplicaAfterSettingTimeout` test on covering `set statement_timeout` use case. But it wasn't tested from this perspective

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
```diff
@@ -47,6 +51,24 @@ public void initialize(Connection connection) throws SQLException {
...
+            return;
+        }
+        try (Statement statement = connection.createStatement()) {
+            for (String parameterChange : runtimeParameterChanges) {
+                statement.execute(parameterChange);
+            }
+        }
+    }
+
+    public void addRuntimeParameterConfiguration(
```
afaruga-atlassian, line 68: for now `runtimeParameterChanges` are only `statements`, do you expect other parameters like this? if yes which?

afaruga-atlassian (approved):
LGTM

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
wyrzyk: I think I don't understand the question. What does it mean `runtimeParameterChanges are only statements`?

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
afaruga-atlassian: Correct, I was wondering about some edge case, but it doesn't make senses as
> internal connections are transparent

👍

src/main/java/com/atlassian/db/replica/internal/ConnectionParameters.java
afaruga-atlassian: this is NIT.
In other words, why `addRuntimeParameterConfiguration` ? 
not `addStatemets`, what do you have in mind?

wyrzyk: `set` sets runtime parameters for the current connection. We add the configurations (in a form of SQL commands), not statements.

PR #150, opened 2022-07-01
Mark release 2.6.4

wyrzyk:


src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicasDiscoverer.java
```diff
@@ -40,9 +38,7 @@ public List<AuroraJdbcUrl> fetchReplicasUrls(Connection connection) throws SQLEx
 
     private List<String> fetchReplicasServerIds(Connection connection) throws SQLException {
         List<String> ids = new LinkedList<>();
-        final String sql = ignoreInactiveReplicas ?
-            "SELECT server_id FROM aurora_replica_status() WHERE session_id != 'MASTER_SESSION_ID' and last_update_timestamp > NOW() - INTERVAL '5 minutes'" :
-            "SELECT server_id FROM aurora_replica_status() WHERE session_id != 'MASTER_SESSION_ID'";
+        final String sql = "SELECT server_id FROM aurora_replica_status() WHERE session_id != 'MASTER_SESSION_ID' and last_update_timestamp > NOW() - INTERVAL '5 minutes'";
```
afaruga-atlassian, line 41: Have you run it on real Aurora ?

wyrzyk: Yes

src/test/java/com/atlassian/db/replica/api/AuroraClusterMock.java
```diff
@@ -52,7 +62,6 @@ public static ResultSet auroraGlobalDbInstanceStatus() {
         SimpleResultSet rs = new SimpleResultSet();
         rs.addColumn("SERVER_ID", Types.VARCHAR, 255, 0);
         rs.addColumn("SESSION_ID", Types.VARCHAR, 255, 0);
-        rs.addRow("apg-global-db-rpo-mammothrw-elephantro-1-n1", "MASTER_SESSION_ID");
```
afaruga-atlassian, line 55: Why this is removed?

afaruga-atlassian (approved):

src/test/java/com/atlassian/db/replica/api/AuroraClusterMock.java
wyrzyk: Before the change, the method returned all database nodes, and the h2 database filtered out MASTER_SESSION_ID. Currently, the whole query is mocked, so this part is no longer needed.

src/main/java/com/atlassian/db/replica/internal/aurora/AuroraReplicasDiscoverer.java
afaruga-atlassian: @wyrzyk Have you thought about extending builder to configure number of seconds/minutes instead having fixed 5 minutes?

wyrzyk: Adding an API is easy, but maintenance is always a cost. We should extend API only with a good use case.

PR #154, opened 2022-09-08
Issue/none lsn not propagated after close

wyrzyk:
I think it may be the issue we observe in the product:
1. We get a connection from the `transaction-thread local`.
2. We prepare a statement and `executeUpdate` without calling `commit`.
3. We close the connection

The transaction is never committed on the DualConnection layer, so we never
register ReplicaConsistency#write, and we don't store the last write.
The following connection assumes the replica is consistent.

Why was the data persistent by the update without calling `commit`?

It's because we have a hook:
```
.addOnClose((rawConnection, timeTaken) -> {
                boolean autocommit = rawConnection.getAutoCommit();
                if (!autocommit) {
                    rawConnection.setAutoCommit(true);
                }
            });
```
The hook calls setAutoCommit(true) if the auto-comit was set to false,
and according to the Connection docs:
"If this method is called during a transaction and the
 auto-commit mode is changed, the transaction is committed.".

### DIFF
--- a/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
+++ b/src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
@@ -175,6 +175,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
         final Optional<Connection> connection = state.getConnection();
         if (connection.isPresent()) {
             connection.get().rollback();
+            state.clearDirty();
         }
     }
 
@@ -191,6 +192,7 @@ public class ReplicaConnectionProvider implements AutoCloseable {
         if (state.getState().equals(MAIN) && !autoCommit) {
             final Connection mainConnection = state.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL));
             consistency.write(mainConnection);
+            state.clearDirty();
         }
     }
 
@@ -200,7 +202,13 @@ public class ReplicaConnectionProvider implements AutoCloseable {
             consistency.preCommit(mainConnection);
         }
     }
+    public void markConnectionDirty(){
+        this.state.markDirty();
+    }
 
+    public boolean isDirty(){
+        return this.state.isDirty();
+    }
     @Override
     public void close() throws SQLException {
         state.close();
--- a/src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
+++ b/src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
@@ -37,6 +37,28 @@ public final class ConnectionState {
     private final DecisionAwareReference<Connection> readConnection;
     private final DecisionAwareReference<Connection> writeConnection;
 
+    /**
+     * When we use a connection to write to the database, it becomes a 'dirty' connection.
+     * The state is cleared when we commit the transaction. Currently, we use it only
+     * to detect dirty connection close() and commit the transaction before closing
+     * the connection.
+     * In the future we can use this state to exit `MainConnection` state and move more traffic to replicas.
+     * (see <a href="https://github.com/atlassian-labs/db-replica/blob/master/docs/dual-connection-states.png">Dual connection states</a>).
+     */
+    private volatile boolean dirty = false;
+
+    public void markDirty() {
+        this.dirty = true;
+    }
+
+    public void clearDirty() {
+        this.dirty = false;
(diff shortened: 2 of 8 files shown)


wyrzyk:
I think the issue is in the product. Transactions should be committed explicitly without relying on a hook's side effects. There is probably more than one place that misbehaves this way. I'll try to fix it on the db-replica side (or et least detect and warn about the issue).

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
```diff
@@ -189,9 +190,16 @@ public void commit() throws SQLException {
     private void recordCommit(boolean autoCommit) throws SQLException {
         if (state.getState().equals(MAIN) && !autoCommit) {
             consistency.write(state.getWriteConnection(new RouteDecisionBuilder(Reason.RW_API_CALL)));
+            state.clearDirty();
```
afaruga-atlassian, line 193: it's cleared only when ` consistency.write` happen, not always-> outside of check? 
I guess then we would do it more excessively.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
```diff
@@ -34,6 +34,27 @@ public final class ConnectionState {
     private final Warnings warnings;
     private final StateListener stateListener;
     private volatile boolean replicaConsistent = true;
+    /**
+     * When we use a connection to write to the database, it becomes a 'dirty' connection.
+     * The state is cleared when we commit the transaction. Currently, we use it only
+     * to detect dirty connection close() and commit the transaction before closing
+     * the connection.
+     * In the future we can use this state to exit `MainConnection` state and move more traffic to replicas.
```
afaruga-atlassian, line 42: This line is interesting, I wonder whether an issue should not be created to don't miss it.
I can't imagine the moment when RR effectiveness it too slow and this would become a priority.

src/main/java/com/atlassian/db/replica/spi/DirtyConnectionCloseHook.java
```diff
@@ -0,0 +1,11 @@
+package com.atlassian.db.replica.spi;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/**
+ * The hook will be invoked before closing a connection with uncommitted data.
+ */
+public interface DirtyConnectionCloseHook {
```
afaruga-atlassian, line 9: UncommittedConnectionCloseHook?

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1334,4 +1337,26 @@ public void shouldClosedConnectionBeNotValid() throws SQLException {
...
+            new PessimisticPropagationConsistency
+                .Builder()
+                .assumeMaxPropagation(Duration.ofDays(10))
+                .cacheLastWrite(lastWriteCache)
+                .build()
+        ).dirtyConnectionCloseHook(Connection::commit)
+            .build();
+        connection.setAutoCommit(false);
+        connection.prepareStatement(SIMPLE_QUERY).executeUpdate();
+//        connection.commit();
```
afaruga-atlassian, line 1357: Do you want to leave it for the sake of visibility? Maybe some comment should be added cause sb might just thought it's a rubbish?

afaruga-atlassian (dismissed):

src/main/java/com/atlassian/db/replica/internal/ReplicaConnectionProvider.java
wyrzyk: Yes `consistency.write` means we committed the changes and we track it with ReplicaConsistency.

src/main/java/com/atlassian/db/replica/internal/state/ConnectionState.java
wyrzyk: At some point, we stopped to optimise the split. This way, the db-connection is less complex.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: Removed

dwatl (commented):
LGTM

dwatl (approved):

PR #157, opened 2022-09-14
Issue/none set connection runtime parameters for uppercase set

wyrzyk:


dwatl (approved):

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
```diff
@@ -38,9 +39,17 @@ public boolean isInsert() {
...
     public boolean isSqlSet() {
+        return compatibleWithPreviousVersion ? isSqlSet_old() : isSqlSet_new();
+    }
+
+    private boolean isSqlSet_old() {
         return sql.startsWith("set");
     }
 
+    private boolean isSqlSet_new() {
+        return sql.startsWith("set") || sql.startsWith("SET");
```
afaruga-atlassian, line 50: how about `sEt` or `SeT` :lol

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1658,6 +1658,43 @@ public void shouldKeepUsingReplicaAfterSettingTimeout() throws SQLException {
...
+    @Test
+    public void shouldKeepUsingReplicaAfterSettingTimeoutUpperCase() throws SQLException {
+        final ConnectionProviderMock connectionProvider = new ConnectionProviderMock();
+        final Connection connection = DualConnection.builder(
+            connectionProvider,
+            permanentConsistency().build()
+        ).build();
+
+        try (Statement statement = connection.createStatement()) {
+            statement.execute("SET statement_timeout to 30000");
```
afaruga-atlassian, line 1670: If this is connected "statment timeouts leaks` then... 🤦 🤦

afaruga-atlassian (approved):

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
wyrzyk: We had the discussion before 😉. Db-replica, in the first place, aims for low overhead, and correctness is in second place. It will not work correctly for `SeT`, but it's not worth reducing the performance of every query to cover this corner case.

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: It might be connected to the leaks if we used an upper case syntax. The other way it may be incorrect are different ways to call a query (executeUpdate, executeQuery etc.). The test case covers only `Statement#execute`.

src/main/java/com/atlassian/db/replica/internal/SqlQuery.java
afaruga-atlassian: > We had the discussion before
Yeah, we had, bad luck that we didn't find SET that time.

PR #158, opened 2022-09-20
JCES-2607: Don't call `DirtyConnectionHook` after rollback

wyrzyk:
We had lots of false positive calls for the scenario.

```
 connection.setAutoCommit(false);
try{ 
connection.prepareStatement(SIMPLE_QUERY).executeUpdate();
}finally{
connection.rollback()
}
...
...
connection.close()
```

### DIFF
--- a/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
+++ b/src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
@@ -13,6 +13,7 @@ import com.atlassian.db.replica.internal.RouteDecisionBuilder;
 import com.atlassian.db.replica.internal.state.State;
 import com.atlassian.db.replica.internal.state.StateListener;
 import com.atlassian.db.replica.spi.DatabaseCall;
+import com.atlassian.db.replica.spi.DirtyConnectionCloseHook;
 import com.atlassian.db.replica.spi.ReplicaConsistency;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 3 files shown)


src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1791,4 +1792,39 @@ public void shouldPropagateLastWrite() throws SQLException {
 
         assertThat(lastWriteCache.get()).isPresent();
     }
+
+    @Test
+    public void shouldNotCallDirtyConnectionHookAfterARollback() throws SQLException {
+        final ConnectionProviderMock provider = getProviderThrowingExceptionsOnConnectionRollback();
+        final DirtyConnectionCloseHook dirtyConnectionCloseHook = mock(DirtyConnectionCloseHook.class);
+        final Connection connection = DualConnection.builder(provider, permanentInconsistency().build())
```
afaruga-atlassian, line 1800: `permanentInconsistency()` does it matter?

wyrzyk: No

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
```diff
@@ -1791,4 +1792,39 @@ public void shouldPropagateLastWrite() throws SQLException {
...
+            .dirtyConnectionCloseHook(dirtyConnectionCloseHook)
+            .build();
+
+        connection.setAutoCommit(false);
```
afaruga-atlassian, line 1804: Wait, shouldn't jira has enabled autoCommit? 
So there are cases when rollback happen, connection is shortly close afterwards and.. what should happen with connection? It shouldn't be committed, that's why you assert on:
```
 verify(dirtyConnectionCloseHook, never()).onClose(any(Connection.class));
```
?
and those false positives... it means that there is no a problem at all, right?

afaruga-atlassian (approved):

src/test/java/com/atlassian/db/replica/api/TestDualConnection.java
wyrzyk: >So there are cases when rollback happen, connection is shortly close afterwards and.. what should happen with connection? It shouldn't be committed

I don't understand it. The transaction was rolled back and the connection closes. Everything seems to work as designed.

PR #164, opened 2022-10-14
Remove circuit breaker related code

wyrzyk:


afaruga-atlassian:
I'm missing an explanation **why** CB is removed.

afaruga-atlassian (approved):
Once **why** will be explained feel free to merge it.

wyrzyk:
> I'm missing an explanation why CB is removed.

It's an internal issue that was never exposed via API or used. Removing unnecessary complexity will make the handover easier.

PR #171, opened 2023-07-20
Mark release 2.9.4

wyrzyk:


fcuozzo:
@wyrzyk was it expected to not merge this PR? I cannot see the 2.9.4 README/CHANGELOG changes in the `master` branch and believe it's because this PR got closed and not merge, is that a correct assumption?

wyrzyk:
We use `compatibleWithPreviousVersion' to safely roll out changes with a feature flag. One of the limitations of this approach is that we can't use a feature flag per change. We use one feature flag per release. Previously, we would roll out the changes just after the release and remove `compatibleWithPreviousVersion` from the code, so that the next release could be rolled out without rolling back the previous fixes. But this time, the rollout of the previous changes hasn't happened yet, so we can't safely remove the compatibility.

PR #177, opened 2023-07-21
Discoverer bug reproduction

wyrzyk:


wyrzyk:
I'm not sure what role the Aurora version plays in this. I was able to get the `discoverer` to stop discovering replicas on 13.9. The `discoverer` worked initially, but stopped discovering the replica after I added another replica (only the first one was visible). After I stopped the first replica, no replicas were detectable.

wyrzyk:
Zrzut ekranu 2023-07-21 o 15 41 35

PR #178, opened 2023-07-24
Changed discovery query to use Longs instead of Integers

bilalshehata:
Using integers caused an error SQLState=22003


fcuozzo (approved):
LGTM

